PR #1, opened 2018-11-02
RFC: Reactive assignments

Rich-Harris:
A proposal for simpler, smaller components with a new approach to reactivity.

[**View formatted RFC**](https://github.com/sveltejs/rfcs/blob/reactive-assignments/text/0001-reactive-assignments.md)

See also the original issue https://github.com/sveltejs/svelte/issues/1826

### DIFF
--- a/text/0000-reactive-assignments.md
+++ b/text/0000-reactive-assignments.md
@@ -456,7 +456,43 @@ Under this proposal, these options would be expressed by a new `<svelte:meta>` t
 
 ### Events
 
-TODO
+Svelte 2 components can fire events with `this.fire(eventName, optionalData)`. These events can be listened to programmatically, with `component.on(eventName, callback)`, or declaratively in component markup:
+
+```html
+<Widget on:flubber="doSomething()">
+```
+
+Since there's no more `this`, there's no more `this.fire`, which means we need to rethink events. This gives us an opportunity to align with web components, where `CustomEvent` is used ([issue here](https://github.com/sveltejs/svelte/issues/1655)) — this also gives us event bubbling, which avoids needing to manually propagate events.
+
+The high-level proposal is to introduce a function called `createEventDispatcher` (🐃) that would return a function for dispatching events:
+
+```js
+import { createEventDispatcher } from 'svelte';
+
+const dispatch = createEventDispatcher();
+
+let i = 0;
+const interval = setInterval(() => {
+  dispatch(i++ % 2 ? 'tick' : 'tock', {
+    answer: 42
+  });
+}, 1000);
+```
+
+This would create a `CustomEvent` with an `event.type` of `tick` (or `tock`) and an `event.detail` of `{ answer: 42 }`.
+
+> 🐃 We could match the arguments to `new CustomEvent(name, opts)` instead — where `detail` is one of the options passed to the constructor alongside things like `bubbles` and `cancelable`
+
+As with lifecycle functions, the `dispatch` is bound to the component because of when `createEventDispatcher` is called.
+
+Listening to events is currently done like so:
+
+```html
+<button on:click="doSomething()">click me</button>
+```
+
+This effectively becomes `{() => this.doSomething()}`. The shorthand is nice, but it is slightly confusing (the context is implicit, and the right-hand side must be a CallExpression which limits flexibility, although there's an [issue for that](https://github.com/sveltejs/svelte/issues/1766)). In the new model, there are no longer any problems around `this`, so it probably makes sense to allow arbitrary expressions instead.
+
 
 ### Component bindings
 
--- a/text/0000-reactive-assignments.md
+++ b/text/0000-reactive-assignments.md
@@ -506,6 +506,8 @@ This would create a `CustomEvent` with an `event.type` of `tick` (or `tock`) and
 
 > 🐃 We could match the arguments to `new CustomEvent(name, opts)` instead — where `detail` is one of the options passed to the constructor alongside things like `bubbles` and `cancelable`
 
+> Note that `new CustomEvent` is unsupported in IE; legacy mode would need to use `document.createEvent('whatever')` instead
+
 As with lifecycle functions, the `dispatch` is bound to the component because of when `createEventDispatcher` is called.
 
 Listening to events is currently done like so:
--- a/text/0000-reactive-assignments.md
+++ b/text/0000-reactive-assignments.md
@@ -867,7 +867,34 @@ Potentially, this could be alleviated with compiler magic (i.e. detecting multip
 
 ### svelte-extras
 
-TODO
+Most of the methods in [svelte-extras](https://github.com/sveltejs/svelte-extras) no longer make sense. The two that we do want to reimplement are `tween` and `spring`.
+
+Happily, these will no longer involve monkey-patching components:
+
+```html
+<script>
+  import { tween } from 'svelte/animation'; // 🐃
+  import * as eases from 'eases-jsnext';
+
+  export let progress = 0;
+  let tweenedProgress = 0;
+
+  let t;
+
+  onstate(() => {
+    if (t) t.stop();
+
+    t = tween(tweenedProgress, progress, {
+      easing: eases.cubicOut,
+      duration: 200
+    }, value => {
+      tweenedProgress = value;
+    });
+  });
+</script>
+
+<progress value={tweenedProgress}/>
+```
 
 
 ### Examples
--- a/text/0000-reactive-assignments.md
+++ b/text/0000-reactive-assignments.md
@@ -622,7 +622,70 @@ The `props` compiler option is redundant now that the contract is defined via `e
 
 ### Dependency tracking
 
-TODO
+In Svelte 2, 'computed properties' use compile-time dependency tracking to derive values from state, avoiding recomputation when dependencies haven't changed. These computed properties are 'push-based' rather than 'pull-based', which can result in unnecessary work:
+
+```html
+{#if visible}
+  <p>{bar}</p>
+{/if}
+
+<script>
+  export default {
+    data: () => ({
+      foo: 1,
+      visible: false
+    }),
+
+    computed: {
+      bar: ({ foo }) => expensivelyCalculateBar(foo)
+    }
+  };
+</script>
+```
+
+In the example above, there is no need to calculate `bar` since it is not rendered.
+
+Under this proposal, there is no longer a separate concept of computed properties. Instead, we can just use functions:
+
+```html
+<script>
+  export let foo = 1;
+  export let visible = false;
+
+  const bar = () => foo;
+</script>
+
+{#if visible}
+  <p>{bar()}</p>
+{/if}
+```
+
+Note that we now *invoke* `bar` in the markup. If `foo` were to change while `visible` were true, we would need to update the contents of `<p>`. That means we need to track the values that could affect the return value, resulting in compiled code similar to this:
+
+```js
+function update(changed, ctx) {
+  if (changed.foo) p.textContent = ctx.bar();
+}
+```
+
+In some situations dependency tracking may be impossible — I'm not sure. In those cases, we can simply bail out:
+
+```js
+function update(changed, ctx) {
+  p.textContent = ctx.bar();
+}
+```
+
+Note that this system is now pull-based rather than push-based. One drawback over the current system of computed properties is that values that are referenced multiple times will also be calculated multiple times:
+
+```html
+{#if visible}
+  <p>{bar()} — {bar()}</p>
+{/if}
+```
+
+Potentially, this could be alleviated with compiler magic (i.e. detecting multiple invocations of pure functions with the same arguments, and computing once before rendering) or userland memoization, but it warrants further exploration.
+
 
 ### Examples
 
--- a/text/0000-reactive-assignments.md
+++ b/text/0000-reactive-assignments.md
@@ -453,6 +453,10 @@ Under this proposal, these options would be expressed by a new `<svelte:meta>` t
 <svelte:meta namespace="svg" tag="my-cool-thing">
 ```
 
+### Script-less components
+
+TODO
+
 
 ### Events
 
@@ -500,7 +504,63 @@ TODO
 
 ### Component API
 
-TODO
+Under this proposal, there is no longer a `this` with state and methods *inside* a component. But there needs to be a way to interact with the component from the outside world.
+
+Instantiating a top-level component probably needn't change (except 🐃 maybe changing `data` to `props`):
+
+```js
+import App from './App.html';
+
+const app = new App({
+  target: document.querySelector('body'),
+  props: {
+    name: 'world'
+  }
+});
+```
+
+Exported properties can be exposed as accessors:
+
+```js
+app.name; // world
+app.name = 'everybody'; triggers a (sync?) update
+```
+
+This creates consistent behaviour between Svelte components that are compiled to custom elements, and those that are not, while also making it easy to understand the component's contract.
+
+Of the five **built-in methods** that currently comprise the [component API](https://svelte.technology/guide#component-api) — `get`, `set`, `fire`, `on` and `destroy` — we no longer need the first three. `on` and `destroy` are still necessary.
+
+In some cases, a component that is designed to be used as a standalone widget will create its own **custom methods**. In Svelte 2, these are lumped in with 'private' (except not really) methods. Under this proposal, custom methods are just exported variables that happen to be functions:
+
+```html
+<script>
+  let visible;
+
+  export function show() {
+    visible = true;
+  }
+
+  export function hide() {
+    visible = false;
+  }
+</script>
+
+<div class="peekaboo">
+  {#if visible}
+    <p>now you see me</p>
+  {/if}
+</div>
+```
+
+```js
+import Peekaboo from './Peekaboo.html';
+
+const peekaboo = new Peekaboo(...);
+peekaboo.show();
+```
+
+An `export const` would be a signal to the compiler that a property is read-only — in other words, attempting to assign to it would cause an error.
+
 
 ### Preload
 
--- a/text/0000-reactive-assignments.md
+++ b/text/0000-reactive-assignments.md
@@ -295,15 +295,15 @@ export function use_interval(fn, ms) {
 
 > This might seem less ergonomic than React Hooks, whereby you can do `const time = useCustomHook()`. The payoff is that you don't need to run that code on every single state change.
 
-There are two other lifecycle functions required — (🐃) `onstate` and (🐃) `onupdate`:
+There are two other lifecycle functions required — (🐃) `onprops` (similar to `onstate` in Svelte v2) and (🐃) `onupdate`:
 
 ```html
 <script>
-  import { onstate, onupdate } from 'svelte';
+  import { onprops, onupdate } from 'svelte';
 
   export let foo;
 
-  onstate(() => {
+  onprops(() => {
     // this callback runs whenever props change
   });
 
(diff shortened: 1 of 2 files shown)


text/0000-reactive-assignments.md
```diff
@@ -0,0 +1,946 @@
...
+Under this proposal, the code inside the `<script>` block *would* run for server-rendered components, which also means that `ondestroy` would need to run. `onstate` and `onupdate` would be no-ops.
+
+I don't think the API needs to change.
+
+
+### Standalone components
+
+It's not clear how this proposal affects standalone components (as opposed to those that use shared helpers). Perhaps it doesn't?
+
+One consequence of the new way of doing things: Svelte no longer lives in `pkg.devDependencies`, but would instead need to become a `dependency`. Not very On Brand but probably a small price to pay.
```
Conduitry: As I noted in chat, I don't see why this is an issue. It seems like we could just inline these `import { whatever } from 'svelte'` statements, and avoid the dependence on `svelte` at runtime

tomcon:
Will equivalents to component.root & component.options be available?

text/0000-reactive-assignments.md
```diff
@@ -0,0 +1,946 @@
...
+Exported properties can be exposed as accessors:
+
+```js
+app.name; // world
+app.name = 'everybody'; triggers a (sync?) update
+```
+
+This creates consistent behaviour between Svelte components that are compiled to custom elements, and those that are not, while also making it easy to understand the component's contract.
+
+Of the five **built-in methods** that currently comprise the [component API](https://svelte.technology/guide#component-api) — `get`, `set`, `fire`, `on` and `destroy` — we no longer need the first three. `on` and `destroy` are still necessary.
```
Conduitry: Do we want to have some naming/prefixing convention for built-in methods (or data, like the current `options` or `root`)? This is a little less easy to type, but it does avoid issues in the future if we want to add something and we need to figure out whether that's a breaking changing change because it might conflict with user-defined items. We can tell users 'don't name anything starting with a `$`' but we can't say 'don't name anything that sounds like we might want to use it later potentially'.

text/0000-reactive-assignments.md
Rich-Harris: yeah fair enough. removed. side-note — I was thinking that it probably makes sense to just use Rollup to create the standalone things (with everything except `svelte` considered external), rather than the current Heath Robinson nonsense

text/0000-reactive-assignments.md
Rich-Harris: Not a bad idea. Vue does `$` prefixes. As long as there's no point of confusion with Store, that's probably a good move

Rich-Harris:
> Will equivalents to component.root & component.options be available?

There's no way to do that *currently* in this RFC. We should consider what the use cases are so that we can figure out if it's necessary, or if there's a better way.

text/0000-reactive-assignments.md
Conduitry: That sounds way saner than the current mechanisms for this, but I am a bit worried about the compiler's size. Am I going to be forced to download Svelte with a bundled-in version of Rollup? Am I going to be forced to install it as a `dependency` even if (for some reason) I don't want it? (Or if I already have a different/newer version of Rollup as a dependency of my app?)

Do we want to make Rollup an optionalDependency, and throw if we try to compile in standalone mode with it unavailable?

text/0000-reactive-assignments.md
```diff
@@ -0,0 +1,944 @@
...
+```
+
+**This behaviour might seem surprising, even shocking at first.** Variable assignment does not have side-effects in JavaScript, meaning that this is arguably something else — [SvelteScript](https://mobile.twitter.com/youyuxi/status/1057291776724271104), perhaps. In practice though, developers readily embrace 'magic' that makes their day-to-day lives easier as long as the mechanisms are ultimately easy to understand — something observed with React Hooks, Vue's reactivity system, Immer's approach to immutable data, and countless other examples. This does underscore the need for this code transformation to be well-documented and explained, however.
+
+
+### Props
+
+Many frameworks (and also web components) have a conceptual distinction between 'props' (values passed *to* a component) and 'state' (values that are internal to the component). Svelte does not. This is a shortcoming.
+
+The proposed solution is to use the `export` keyword to declare a value that can be set from outside as a prop:
```
Rich-Harris: One piece of feedback I've received objected to using `export` because 'props flow in, not out' — this feedback was accompanied by an analogy I won't repeat here.

It's not strictly accurate, since this RFC proposes that properties be attached to the component instances as accessors, as happens with web components, and it is a means by which we can specify a custom component API. Nevertheless, people are likely to react similarly, so we should perhaps adjust the language around this to discuss alternatives and explain why they're unsuitable. (The suggestion here was to use a magic `import`, but aside from the inherent problems with magic imports, that wouldn't allow us to specify default values.)

text/0000-reactive-assignments.md
```diff
@@ -456,7 +456,43 @@ Under this proposal, these options would be expressed by a new `<svelte:meta>` t
...
+const interval = setInterval(() => {
+  dispatch(i++ % 2 ? 'tick' : 'tock', {
+    answer: 42
+  });
+}, 1000);
+```
+
+This would create a `CustomEvent` with an `event.type` of `tick` (or `tock`) and an `event.detail` of `{ answer: 42 }`.
+
+> 🐃 We could match the arguments to `new CustomEvent(name, opts)` instead — where `detail` is one of the options passed to the constructor alongside things like `bubbles` and `cancelable`
```
lukeed: If we are to use `new CustomEvent`, we are dropping IE9-11 support. Not sure that this is an intended consequence. This would have to be this in order to accomodate:

```js
let ev = document.createEvent('CustomEvent');
ev.initCustomEvent(...);
```

TehShrike:
The `svelte:meta` element seems like a great way to get the component's state as an object, or even just get the props that were passed in:

```html
<svelte:meta bind:state bind:props bind:options bind:root=parent />

<script>
const cool = true
console.log(state.cool)
</script>
```

My original thought was to use `ref` instead of `bind`, but `bind` seems better.  @Conduitry noted that already uses the `bind` directive for one-way binding, so it doesn't seem too weird here.

text/0000-reactive-assignments.md
Rich-Harris: definitely a fair concern. also, it would force `svelte.compile` to become async. maybe we should cross that bridge later

text/0000-reactive-assignments.md
Rich-Harris: ah thanks, didn't realise. That could done via `legacy: true`

text/0000-reactive-assignments.md
lukeed: I don't understand why `fire` and `on` are removed?

A method's sole purpose may be to emit an event, and rather than pass the Component a `props.onFoobar` , something else can listen to Component for that sequence

```js
const greet = new Greeter();
const stutter = str => str.charAt(0).repeat(3) + str;

greet.name = 'Bob';
greet.name = 'George';

greet.on('pretty-girl', stutter);
```

Rich-Harris:
Yes, I like the `<svelte:meta>` idea. Idiomatic, easy to compile. 

I don't think we'd want it to make those names magically available in `<script>` though — I think we'd want to handle it the same way this RFC proposes we handle refs:

```html
<script>
  import Widget from './Widget.html';

  let props;
  let options;
  let canvas;

  onmount(() => {
    console.log(props, options, canvas);
  });

  onprops(() => {
    console.log(props);
  });
</script>

<svelte:meta bind:props bind:options/>

<canvas ref:canvas></canvas>
<Widget {...props}/>
```

We probably don't want a binding for `state`. `props` makes sense, as does `root`. Not entirely sold on `options` — it'd be nice if we were free to use a more optimal private API for inline components, rather than the public API (which would also prevent us from doing certain optimisations like component folding).

TehShrike:
Oh, I'd somehow missed that bit about needing to define the variable with `let` for refs.

Yeah, that would make TypeScript type waaaaaay simpler.

text/0000-reactive-assignments.md
```diff
@@ -598,7 +598,20 @@ TODO
 
 ### TypeScript
 
-TODO
+A major advantage of this proposal over Svelte 2 is that components become far more amenable to typechecking and other forms of static analysis. Even without TypeScript, VSCode is able to offer much richer feedback now than with Svelte 2.
+
+An eventual goal is to be able to use TypeScript in components:
+
+```html
+<script type="typescript">
```
lukeed: `<script lang="typescript">` imo – `type` should be reserved for `type="module"` (native).

text/0000-reactive-assignments.md
```diff
@@ -607,9 +607,46 @@ peekaboo.show();
...
+```
+
+This is deeply weird, and due to an oversight (that we can't correct without a breaking change) only runs on the client.
+
+Since Sapper requires that components have some way of declaring their data dependencies prior to rendering, and since `setup` is so cumbersome, there is a special case made for `preload`. The `preload` function is attached to components on both client and server, and has no well-defined behaviour; it is purely convention.
+
+We can do better, **but it requires something potentially controversial** — a second `<script>` block that, unlike the one we've been using so far, runs a single time rather than upon every instantiation. Let's call it (🐃) `scope="shared"`:
+
+```html
+<script scope="shared">
```
lukeed: As mentioned in chat, I would _really_ like to see this be written as this instead:

```html
<!-- ambiguous "preload" applies to both scopes -->
<script preload>
  export function preload({ params }) {
    //
  }
</script>

<!-- specify "preload" scope -->
<script preload="client|server">
  export function preload({ params }) {
    //
  }
</script>
```

text/0000-reactive-assignments.md
```diff
@@ -650,7 +650,14 @@ We can do better, **but it requires something potentially controversial** — a
 
 ### Server-side rendering
 
-TODO
+The compiler running with the `generate: 'ssr'` option produces completely different code from `generate: 'dom'`. This is because server-rendered components do not have a lifespan — they are rendered, then immediately discarded, so rather than having a concept of 'instances' an SSR component exports a function that given some data creates HTML and CSS.
+
+This does create some subtle behaviour differences however: there is no real place to do any kind of setup work, and `oncreate` (and the other lifecycle hooks) never run. Component bindings are also brittle.
```
lukeed: I think it's important to clarify here – if `oncreate` is to be renamed to `onmount` (aka, the component is attached to DOM now), then yes, it has no business being invoked in SSR context.

Only mention it because we briefly mentioned renamed `oncreate` to `onmount` (to serve as a clearer equivalent to `componentDidMount`). Then, perhaps, a _new_ `oncreate` would be added, serving as the `componentWillMount` equivalent – would _would_ make it redundant & useless in this RFC.

text/0000-reactive-assignments.md
Conduitry: Do we want to _drop_ standalone component compilation as part of `svelte.compile` and introduce a separate function that does require Rollup, or maybe some sort of drop-in Rollup plugin or config?

One possibility would be to make this not `svelte.compile`'s concern at all, and to take care of it in the `component-template` `rollup.config.js`.

I do see the value of no longer having to worry about this weirdness, but I am concerned that we will be stuck with it until Svelte v4 unless we do _something_ as part of these changes.

text/0000-reactive-assignments.md
```diff
@@ -666,6 +658,73 @@ It's not clear how this proposal affects standalone components (as opposed to th
...
+The store can also be accessed programmatically:
+
+```js
+const { foo } = this.store.get();
+this.store.doSomething();
+```
+
+Since there is no longer a `this`, we need to reconsider this approach. At the same time, the `get`/`set`/`on`/`fire` interface (designed to mirror the component API) feels slightly anachronistic in light of the rest of this RFC. A major problem with the store in its current incarnation is its lack of typechecker-friendliness, and the undocumented hoops necessary to jump through to integrate with popular libraries like MobX.
+
+I'm not yet sure how the store fits into this proposal — this is perhaps the biggest unknown at this point.
```
lukeed: Not sure if I see the problem?

```html
<h1>Hello, {{name}}!</h1>

<script>
  import { ondestroy } from 'svelte';
  import { Store } from './my-store';
  
  // setup code
  export let name = 'world';
  const store = new Store({ name });

  const listener = store.on('state', ({ current }) => {
    // something changed... changes this component's "name"
    name = Math.random().toString(36).slice(7);
    // this component now updates
  });

  // cleanup
  ondestroy(listener.cancel);

File: text/0000-reactive-assignments.md (hunk 6 above)
Comment (Conduitry) [reply text/0000-reactive-assignments.md]: I actually think the unification of the different types of static stuff that can be attached to a component is nice - and being able to make `preload` just be a convention is great. (If we want to have different client/server preloads, could that just be part of Sapper's convention?)

That said, I am really not fond of the term `scope='shared'` that's being proposed to indicate this. I don't have a better suggestion currently.

File: text/0000-reactive-assignments.md [hunk 9]
```diff
@@ -666,6 +658,73 @@ It's not clear how this proposal affects standalone components (as opposed to th
...
+    const { thingIDontWant, ...everythingElse } = __props__;
+    return everythingElse;
+  };
+</script>
+
+<Foo {...__props__}/>
+<Bar {...subset()}/>
+```
+
+This is inelegant, but I believe the inelegance would affect a small minority of components.
```
lukeed: If we're using special prefix `$` like @Conduitry suggested, this could just be `$props` which _looks_ 10x better.

```html
<Foo {...$props} />

<script>
  import Foo from './Foo.html';
  import Bar from './Bar.html';
    const subset = () => {
    const { thingIDontWant, ...everythingElse } = $props;
    return everythingElse;
  };
</script>
```

text/0000-reactive-assignments.md
Conduitry: I think what that loses is being able to have the store Just Be There without having to manually import it in every component, with the different relative paths that will entail.

Perhaps we could use @TehShrike's `<svelte:meta bind:XXX/>` idea for this?

I'm not certain we want to maintain a magic way of referring to store properties in templates (the `$`-prefixed identifiers).

text/0000-reactive-assignments.md
Conduitry: My idea with the `$` prefix was that it was for externally-visible methods/data on the component instance. For making stuff available to code inside the component, I'm currently favoring something like @TehShrike's `<svelte:meta bind:XXX/>` idea of magic globals.

text/0000-reactive-assignments.md
lukeed: ^ My concerns erased. Solved in chat.

text/0000-reactive-assignments.md
lukeed: Personally I'd like to get rid of magical store presence. Not everything _needs_ to connect to a/the store or have access to it.

Doing it this way also lets a develop to define a custom top-level store (or multiple) and import the ones they want within any given component.

As for `$vars` being accessible – I think that can still be achieved if we wanted to. If the compiler can link up `foobar` it should be able to link up `$foobar` too?

But, personally, I'd prefer to do away with magic `$vars` and have the store update the variables I've declared & wired up during setup. For example, the `export let name` being updated within store listener.

tivac: > But, personally, I'd prefer to do away with magic $vars and have the store update the variables I've declared & wired up during setup. For example, the export let name being updated within store listener.

Strong disagree on this sentiment from me. We're using stores heavily and the ability for children to inherit parent stores and transparently reference `$vars` is super valuable to our sanity and iteration speed.

lukeed: Fair 😆 For context, I use store in less than 10% of my components.

text/0000-reactive-assignments.md
Rich-Harris: I think we should strongly consider it. I don't think it needs to make it into *this* RFC (it doesn't need to cover every aspect of v3) but yeah, we should try not to lose track of it. Standalone components are a useful selling point of this framework but they are also a total bloody nuisance

text/0000-reactive-assignments.md
Rich-Harris: Updated the text, resolving this item

text/0000-reactive-assignments.md
Rich-Harris: updated

text/0000-reactive-assignments.md
Rich-Harris: I'm inclined to agree with @Conduitry — `preload` is an awkward anomaly. The less we have those special cases the more stable a platform for innovation the framework becomes

text/0000-reactive-assignments.md
Rich-Harris: There is arguably room for a `componentWillMount` equivalent, since there's currently nothing that runs a single time between the first time props are injected and the first time the component renders. It *is* achievable in userland using the `onprops` (are we calling it that?) function though.

👍 to `onmount`

text/0000-reactive-assignments.md
Conduitry: Wait I just realized, this would also probably let us leave behind all of the non-ESM output modes too. Dang this is sounding real tasty now.

Conduitry: Ugh I just remembered the Svelte CLI which people are still using for some reason. I don't know how that fits into this anymore. Do we tell the people who were using it to create standalone IIFEs to get lost?

text/0000-reactive-assignments.md
lukeed: TBH I'm not sure what either of you are saying 😅

Rich-Harris: There's nothing special about `preload` — it doesn't have any defined behaviour. It's just a function that will get attached to both client and server components as a static method, and it gets that privileged treatment because Sapper needs it.

But ideally Svelte shouldn't have features that exist just for the sake of Sapper — someone with different opinions about how to build an app framework with Svelte should be able to do so. 

This RFC, among other things, is about stripping Svelte down to its bare minimum — two methods instead of five, one way of making values available to the markup instead of (3? 10? depends which ones you count).

Since we need some way of including code that runs once per app (instead of once per component), we need to have the second `<script>` tag. Since we have that second script tag, we may as well use it for `preload` rather than it being its own special thing.

lukeed: 100% agreed - apologies if something I said pointed to the contrary. I was merely just renaming `scope` 

Having extra script tags also _allows_ new app frameworks to define new boundaries for new behaviors, especially if/when Svelte users are accustomed to having them already.

Rich-Harris: Ah ok, I thought you were implying that the second `<script>` was purely for `preload`. Certainly in the market for alternatives to `scope`

lukeed: No, but in Sapper apps that would/could be the place for it... Standlone, hoistable preload functions that the Svelte compiler can rip out (safely) per-bundle-type and put those snippets wherever Sapper tells it to put them.

Svelte just needs to be aware of/handle more than 1 script tag

Rich-Harris:
Sorry everyone, I did a bad git. I've opened a new PR for this RFC — #4. I think most of the comments in this thread have been addressed in any case; the big TODO is Store, which we can discuss over there.

text/0000-reactive-assignments.md
tomcon: I'm with @tivac on this one.
The (relatively) simple & consistent (ie. same api) implementation of Store in v2 won a lot of people over imo - including me! By all means knock yourself out with immer etc  etc but in some parts of the real world it is ***very*** useful indeed

PR #2, opened 2018-11-02
Fixing minor syntax issue in example

TehShrike:
Assuming that the function is supposed to return an object


Conduitry:
I would have just merged this in if I wasn't worried about it throwing off Rich when he went to push more commits to a branch that had changed. Maybe for tiny things like this use the new \`\`\`suggestion feature in PR comments?

TehShrike:
Oh yeah, I forgot about that new feature

Josh

________________________________

PR #4, opened 2018-11-03
Reactive assignments

Rich-Harris:
A proposal for simpler, smaller components with a new approach to reactivity.

[**View formatted RFC**](https://github.com/sveltejs/rfcs/blob/reactive-assignments/text/0001-reactive-assignments.md)

See also the original issue https://github.com/sveltejs/svelte/issues/1826

---

*Due to an unfortunate incident at the commit factory, the original PR for this RFC was merged prematurely. Earlier comments can be read [there](https://github.com/sveltejs/rfcs/pull/1)*

### DIFF
--- a/text/0001-reactive-assignments.md
+++ b/text/0001-reactive-assignments.md
@@ -24,7 +24,7 @@ Many developers report being intrigued by Svelte, but unwilling to accommodate s
   export default {
     components: { Nested },
     transitions: { fade }
-  }
+  };
 </script>
 ```
 
@@ -42,7 +42,7 @@ Declaring methods is also somewhat cumbersome, and in many editor configurations
         const { foo } = this.get();
       }
     }
-  }
+  };
 </script>
 ```
 
--- a/text/0001-reactive-assignments.md
+++ b/text/0001-reactive-assignments.md
@@ -1,6 +1,6 @@
 - Start Date: 2018-11-01
-- RFC PR: (leave this empty)
-- Svelte Issue: (leave this empty)
+- RFC PR: https://github.com/sveltejs/rfcs/pull/1
+- Svelte Issue: https://github.com/sveltejs/svelte/issues/1826
 
 # Reactive assignments
 
--- a/text/0000-reactive-assignments.md
+++ b/text/0000-reactive-assignments.md
@@ -456,7 +456,43 @@ Under this proposal, these options would be expressed by a new `<svelte:meta>` t
 
 ### Events
 
-TODO
+Svelte 2 components can fire events with `this.fire(eventName, optionalData)`. These events can be listened to programmatically, with `component.on(eventName, callback)`, or declaratively in component markup:
+
+```html
+<Widget on:flubber="doSomething()">
+```
+
+Since there's no more `this`, there's no more `this.fire`, which means we need to rethink events. This gives us an opportunity to align with web components, where `CustomEvent` is used ([issue here](https://github.com/sveltejs/svelte/issues/1655)) — this also gives us event bubbling, which avoids needing to manually propagate events.
+
+The high-level proposal is to introduce a function called `createEventDispatcher` (🐃) that would return a function for dispatching events:
+
+```js
+import { createEventDispatcher } from 'svelte';
+
+const dispatch = createEventDispatcher();
+
+let i = 0;
+const interval = setInterval(() => {
+  dispatch(i++ % 2 ? 'tick' : 'tock', {
+    answer: 42
+  });
+}, 1000);
+```
+
+This would create a `CustomEvent` with an `event.type` of `tick` (or `tock`) and an `event.detail` of `{ answer: 42 }`.
+
+> 🐃 We could match the arguments to `new CustomEvent(name, opts)` instead — where `detail` is one of the options passed to the constructor alongside things like `bubbles` and `cancelable`
+
+As with lifecycle functions, the `dispatch` is bound to the component because of when `createEventDispatcher` is called.
+
+Listening to events is currently done like so:
+
+```html
+<button on:click="doSomething()">click me</button>
+```
+
+This effectively becomes `{() => this.doSomething()}`. The shorthand is nice, but it is slightly confusing (the context is implicit, and the right-hand side must be a CallExpression which limits flexibility, although there's an [issue for that](https://github.com/sveltejs/svelte/issues/1766)). In the new model, there are no longer any problems around `this`, so it probably makes sense to allow arbitrary expressions instead.
+
 
 ### Component bindings
 
--- a/text/0000-reactive-assignments.md
+++ b/text/0000-reactive-assignments.md
@@ -506,6 +506,8 @@ This would create a `CustomEvent` with an `event.type` of `tick` (or `tock`) and
 
 > 🐃 We could match the arguments to `new CustomEvent(name, opts)` instead — where `detail` is one of the options passed to the constructor alongside things like `bubbles` and `cancelable`
 
+> Note that `new CustomEvent` is unsupported in IE; legacy mode would need to use `document.createEvent('whatever')` instead
+
 As with lifecycle functions, the `dispatch` is bound to the component because of when `createEventDispatcher` is called.
 
 Listening to events is currently done like so:
--- a/text/0000-reactive-assignments.md
+++ b/text/0000-reactive-assignments.md
@@ -867,7 +867,34 @@ Potentially, this could be alleviated with compiler magic (i.e. detecting multip
 
 ### svelte-extras
 
-TODO
+Most of the methods in [svelte-extras](https://github.com/sveltejs/svelte-extras) no longer make sense. The two that we do want to reimplement are `tween` and `spring`.
+
+Happily, these will no longer involve monkey-patching components:
+
+```html
+<script>
+  import { tween } from 'svelte/animation'; // 🐃
+  import * as eases from 'eases-jsnext';
+
+  export let progress = 0;
+  let tweenedProgress = 0;
+
+  let t;
+
+  onstate(() => {
+    if (t) t.stop();
+
+    t = tween(tweenedProgress, progress, {
+      easing: eases.cubicOut,
+      duration: 200
+    }, value => {
+      tweenedProgress = value;
+    });
+  });
+</script>
+
+<progress value={tweenedProgress}/>
+```
 
 
 ### Examples
--- a/text/0000-reactive-assignments.md
+++ b/text/0000-reactive-assignments.md
@@ -622,7 +622,70 @@ The `props` compiler option is redundant now that the contract is defined via `e
 
 ### Dependency tracking
 
-TODO
+In Svelte 2, 'computed properties' use compile-time dependency tracking to derive values from state, avoiding recomputation when dependencies haven't changed. These computed properties are 'push-based' rather than 'pull-based', which can result in unnecessary work:
+
+```html
+{#if visible}
+  <p>{bar}</p>
+{/if}
+
+<script>
+  export default {
+    data: () => ({
+      foo: 1,
+      visible: false
+    }),
+
+    computed: {
+      bar: ({ foo }) => expensivelyCalculateBar(foo)
+    }
+  };
+</script>
+```
+
+In the example above, there is no need to calculate `bar` since it is not rendered.
+
+Under this proposal, there is no longer a separate concept of computed properties. Instead, we can just use functions:
+
+```html
+<script>
+  export let foo = 1;
+  export let visible = false;
+
+  const bar = () => foo;
+</script>
+
+{#if visible}
+  <p>{bar()}</p>
+{/if}
+```
+
+Note that we now *invoke* `bar` in the markup. If `foo` were to change while `visible` were true, we would need to update the contents of `<p>`. That means we need to track the values that could affect the return value, resulting in compiled code similar to this:
+
+```js
+function update(changed, ctx) {
+  if (changed.foo) p.textContent = ctx.bar();
+}
+```
+
+In some situations dependency tracking may be impossible — I'm not sure. In those cases, we can simply bail out:
+
+```js
+function update(changed, ctx) {
+  p.textContent = ctx.bar();
+}
+```
+
+Note that this system is now pull-based rather than push-based. One drawback over the current system of computed properties is that values that are referenced multiple times will also be calculated multiple times:
+
+```html
+{#if visible}
+  <p>{bar()} — {bar()}</p>
+{/if}
+```
+
+Potentially, this could be alleviated with compiler magic (i.e. detecting multiple invocations of pure functions with the same arguments, and computing once before rendering) or userland memoization, but it warrants further exploration.
+
 
 ### Examples
 
--- a/text/0000-reactive-assignments.md
+++ b/text/0000-reactive-assignments.md
@@ -453,6 +453,10 @@ Under this proposal, these options would be expressed by a new `<svelte:meta>` t
 <svelte:meta namespace="svg" tag="my-cool-thing">
 ```
 
+### Script-less components
+
+TODO
+
 
 ### Events
 
@@ -500,7 +504,63 @@ TODO
 
 ### Component API
 
-TODO
+Under this proposal, there is no longer a `this` with state and methods *inside* a component. But there needs to be a way to interact with the component from the outside world.
+
+Instantiating a top-level component probably needn't change (except 🐃 maybe changing `data` to `props`):
+
+```js
+import App from './App.html';
+
+const app = new App({
+  target: document.querySelector('body'),
+  props: {
+    name: 'world'
+  }
+});
+```
+
+Exported properties can be exposed as accessors:
+
+```js
+app.name; // world
+app.name = 'everybody'; triggers a (sync?) update
+```
+
+This creates consistent behaviour between Svelte components that are compiled to custom elements, and those that are not, while also making it easy to understand the component's contract.
+
+Of the five **built-in methods** that currently comprise the [component API](https://svelte.technology/guide#component-api) — `get`, `set`, `fire`, `on` and `destroy` — we no longer need the first three. `on` and `destroy` are still necessary.
+
+In some cases, a component that is designed to be used as a standalone widget will create its own **custom methods**. In Svelte 2, these are lumped in with 'private' (except not really) methods. Under this proposal, custom methods are just exported variables that happen to be functions:
+
+```html
+<script>
+  let visible;
+
+  export function show() {
+    visible = true;
+  }
+
+  export function hide() {
+    visible = false;
+  }
+</script>
+
+<div class="peekaboo">
+  {#if visible}
+    <p>now you see me</p>
+  {/if}
+</div>
+```
+
+```js
+import Peekaboo from './Peekaboo.html';
+
+const peekaboo = new Peekaboo(...);
+peekaboo.show();
+```
+
+An `export const` would be a signal to the compiler that a property is read-only — in other words, attempting to assign to it would cause an error.
+
 
 ### Preload
 
--- a/text/0000-reactive-assignments.md
+++ b/text/0000-reactive-assignments.md
@@ -295,15 +295,15 @@ export function use_interval(fn, ms) {
 
 > This might seem less ergonomic than React Hooks, whereby you can do `const time = useCustomHook()`. The payoff is that you don't need to run that code on every single state change.
 
-There are two other lifecycle functions required — (🐃) `onstate` and (🐃) `onupdate`:
+There are two other lifecycle functions required — (🐃) `onprops` (similar to `onstate` in Svelte v2) and (🐃) `onupdate`:
 
 ```html
 <script>
-  import { onstate, onupdate } from 'svelte';
+  import { onprops, onupdate } from 'svelte';
 
   export let foo;
 
-  onstate(() => {
+  onprops(() => {
     // this callback runs whenever props change
   });
 
@@ -340,17 +340,17 @@ Currently, `onstate` and `onupdate` run before and after every state change. Bec
 </script>
 ```
 
-Under this proposal, the `onstate` callback runs whenever props change but *not* when private state changes. This makes cycles impossible:
+Under this proposal, the `onprops` callback runs whenever props change but *not* when private state changes. This makes cycles impossible:
 
 ```html
 <script>
-  import { onstate } from 'svelte';
+  import { onprops } from 'svelte';
 
   export let temperature = 32;
   let getting = null;
   let previous_temperature;
 
-  onstate(() => {
+  onprops(() => {
     // this assignment will *not* result in the callback
     // firing again
     getting = temperature > previous_temperature
@@ -369,13 +369,13 @@ Any `onupdate` callbacks would run after the view was updated, whether as a resu
 As previously mentioned, `oncreate` is used in Svelte 2 to run code after the initial render has taken place. Strictly speaking it is redundant...
 
 ```js
-import { onstate, onupdate } from 'svelte';
+import { onprops, onupdate } from 'svelte';
 import { once } from 'lodash-es';
 
 export let externalProp;
 let internalProp = 'defined';
 
-onstate(once(() => {
+onprops(once(() => {
   // externalProp is now defined
 }));
 
@@ -384,7 +384,7 @@ onupdate(once(() => {
 }));
 ```
 
-...but we could decide (🐃) to include `oncreate` as a convenience anyway.
+...but we could decide (🐃) to include an `oncreate` equivalent as a convenience anyway. Currently, the favoured name for this function is `onmount`.
 
 
 ---
@@ -661,7 +661,7 @@ The compiler running with the `generate: 'ssr'` option produces completely diffe
 
 This does create some subtle behaviour differences however: there is no real place to do any kind of setup work, and `oncreate` (and the other lifecycle hooks) never run. Component bindings are also brittle.
 
-Under this proposal, the code inside the `<script>` block *would* run for server-rendered components, which also means that `ondestroy` would need to run. `onstate` and `onupdate` would be no-ops.
+Under this proposal, the code inside the `<script>` block *would* run for server-rendered components, which also means that `ondestroy` would need to run. `onprops` and `onupdate` would be no-ops.
 
 I don't think the API needs to change.
 
@@ -894,7 +894,7 @@ Happily, these will no longer involve monkey-patching components:
 
   let t;
 
-  onstate(() => {
+  onprops(() => {
     if (t) t.stop();
 
     t = tween(tweenedProgress, progress, {
--- a/text/0000-reactive-assignments.md
+++ b/text/0000-reactive-assignments.md
@@ -379,6 +379,11 @@ onupdate(once(() => {
 ...but we could decide (🐃) to include `oncreate` as a convenience anyway.
 
 
+---
+
+The remainder of this section will address how existing Svelte concepts are affected by this RFC. Some things (actions, transitions etc) are unmentioned because they are unaffected, though if items are missing please raise your voice.
+
+
 ### Refs
 
 Refs are references to DOM nodes:
--- a/text/0000-reactive-assignments.md
+++ b/text/0000-reactive-assignments.md
@@ -718,24 +718,26 @@ It's convenient to be able to pass down props from a parent to a child component
 </script>
 ```
 
-That opportunity no longer exists in this RFC. Instead we need some other way to express the concept of 'all the properties that were passed into this component'. The best I can come up with is an injected variable with a name like `__props__`:
+That opportunity no longer exists in this RFC. Instead we need some other way to express the concept of 'all the properties that were passed into this component'. One suggestion is to use the `bind` directive on (🐃) the `<svelte:meta>` element described above:
 
 ```html
+<svelte:meta bind:props/>
+
 <script>
   import Foo from './Foo.html';
   import Bar from './Bar.html';
 
+  let props;
+
   const subset = () => {
-    const { thingIDontWant, ...everythingElse } = __props__;
+    const { thingIDontWant, ...everythingElse } = props;
     return everythingElse;
   };
 </script>
 
-<Foo {...__props__}/>
+<Foo {...props}/>
 <Bar {...subset()}/>
 ```
-
-This is inelegant, but I believe the inelegance would affect a small minority of components.
 
 > Since a component's public API ordinarily uses accessors to define the component's contract, it is likely that *inline* components would need to use a privileged non-public API, so that unanticipated props could be passed in. (This is something we should do anyway)
 
--- a/text/0000-reactive-assignments.md
+++ b/text/0000-reactive-assignments.md
@@ -899,7 +899,13 @@ Happily, these will no longer involve monkey-patching components:
 
 ### Examples
 
-TODO
+* markdown editor [current](https://svelte.technology/repl?version=2.15.0&demo=binding-textarea) / [imagined](https://svelte.technology/repl?version=2.15.0&gist=a0443aa0fc68947b5fad8fae1aa63627)
+* media elements [current](https://svelte.technology/repl?version=2.15.0&demo=binding-media-elements) / [imagined](https://svelte.technology/repl?version=2.15.0&gist=8562a70e1a1e708b735b7a50e80b3cfe)
+* nested components [current](https://svelte.technology/repl?version=2.15.0&demo=nested-components) / [imagined](https://svelte.technology/repl?version=2.15.0&gist=cbad006c2197633ac75fc8f31c3670df)
+* SVG clock [current](https://svelte.technology/repl?version=2.15.0&demo=svg-clock) / [imagined](https://svelte.technology/repl?version=2.15.0&gist=758aacc71f506518a70c0e42f0735f6c)
+* Simple transition [current](https://svelte.technology/repl?version=2.15.0&demo=transitions-fade) / [imagined](https://svelte.technology/repl?version=2.15.0&gist=7bd7b4349a27342ca96b58046eb6e765)
+* Custom transition [current](https://svelte.technology/repl?version=2.15.0&demo=transitions-custom) / [imagined](https://svelte.technology/repl?version=2.15.0&gist=2e9830aad338305877d226d80a9d04d4)
+* Temperature converter [current](https://svelte.technology/repl?version=2.15.0&demo=7guis-temperature) / [imagined](https://svelte.technology/repl?version=2.15.0&gist=439fc5e7c89d91a31c9c9c1447434532)
 
 
 ## How we teach this
--- a/text/0000-reactive-assignments.md
+++ b/text/0000-reactive-assignments.md
@@ -607,9 +607,46 @@ peekaboo.show();
 An `export const` would be a signal to the compiler that a property is read-only — in other words, attempting to assign to it would cause an error.
 
 
-### Preload
+### `preload` and `setup`
 
-TODO
+There is a rather awkward mechanism for declaring static properties on a component constructor in Svelte 2 — the `setup` hook:
+
+```html
+<script>
+  export default {
+    setup(Component) {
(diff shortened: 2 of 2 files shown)


timhall:
Looking great! I would say the biggest confusion / stumbling block for me has been the mix of `{...}` and strings in templates and this looks like a great time to get rid of the special string cases, which I think are all directives:

__Event handlers__

```html
<p>Count: {count}</p>
<button on:click={() => count += 1}>+1</button>
```

__Component events__

_Note:_ I don't really see the need for `createEventDispatcher`, seems like something the compiler can sort out.

```html
<script>
  import { dispatch as fire } from 'svelte';
</script>

<p>Select a category:</p>

{#each categories as category}
  <button on:click={() => fire('select', { category })}>select {category}</button>
{/each}

```

__Refs__

```html
<script>
  let canvas;
</script>

<canvas ref={canvas}></canvas>
```

__Transitions__

```html
<script>
  import { fade, fly } from 'svelte-transitions';
</script>

{#if visible}
  <p in:{fly}={{ y: 50 }} out:{fade}>fades in and out</p>
{/if}
```

__Bindings__

```html
<Widget bind:childValue={parentValue} />

<Widget bind:{value} />
```

__Actions__

```html
<button on:click="toggleLanguage" use:{tooltip}={translations[language].tooltip}>
  {language}
</button>
```

__Classes__

```html
<ul class="links">
  <li class:active={url === '/'}>...</li>
  <li class:active={url.startsWith('/blog')}>...</li>
  <li class:active={url.startsWith('/about')}>...</li>
</ul>
```

It's a small thing in some cases, but I think consistency will be a big win overall and make clear that if you need to access / evaluate js in templates, use `{ ... }` tags

timhall:
As for an approach to `store`, I think leaning on reactive assignments would work very well:

```html
<script>
  import { store } from 'svelte';
  
  let { online = navigator.onLine } = store();
  window.addEventListener('offline', () => online = false);
  window.addEventListener('online', () => online = true);
</script>

<p>You are {online ? 'online' : 'offline'}</p>
```

Compiles roughly to:

```html
<script>
import { store } from 'svelte';

const Component = defineComponent((__update, __props) => {
  // (like lifecycle function, get store for current instance)
  const __store = store();

  // This part is a little icky, may be a better way to do this
  let { online } = __store.get();
  if (online === void 0) {
    online = navigator.onLine;
    __store.set({ online });
  }

  __store.on('state', values => {
    online = values.online;
    __update({ online: true });
  });
  window.addEventListener('offline', () => {
    online = false;
    __store.set({ online });
  });
  window.addEventListener('online', () => {
    online = true;
    __store.set({ online });
  });
}, create_main_fragment);
</script>
```

The above should satisfy the current store needs, although there is no built-in `$...` in templates, which is a bit of a shame (although arguably less magic). This could be take a step further and generalized to support component context and the idea of connecting to observables / subjects = `Context` (🐃). 

```ts
interface Context<T> {
  get(): T;
  set(T): void;
  subscribe(callback: (T) => void): () => void
}
```

```html
<script>
  import CustomStore from './custom-store';
  import { connect, context } from 'svelte';

  let { online = navigator.onLine } = connect(context(CustomStore));

  // connect(context(...)) is most likely redundant, context(...) should be enough
  // but connect(...) is nice for values not from context (e.g. Observables)
  
  window.addEventListener('offline', () => online = false);
  window.addEventListener('online', () => online = true);
</script>
```

Compiles roughly to:

```html
<script>
import CustomStore from './custom-store'; 
import { context, ondestroy } from 'svelte';

const Component = defineComponent((__update, __props) => {
  let online = navigator.onLine;

  // (like lifecycle function to find value in current instance's context)
  const __CustomStoreContext = context(CustomStore);
 
  let __values = __CustomStoreContext.get();
  let { online } = __values;
  if (online === void 0) {
    online = navigator.onLine;
    __CustomStoreContext.set({ ...__values, online });
  }

  const unsubscribe = __CustomStoreContext.subscribe(values => {
    online = values.online;
    __update({ online: true });
  });
  ondestroy(unsubscribe);

  window.addEventListener('offline', () => {
    online = false;
    __CustomStoreContext.set({ online });
  });
  window.addEventListener('online', () => {
    online = true;
    __CustomStoreContext.set({ online });
  });
}, create_main_fragment);
</script>
```

Maybe this is a bit too much power (mainly the `connect` bit), but it seems to expand into a fairly minimal compiled output and would make it much easier to pass around observable values (especially if `props` and `store` were built on this primitive).

arxpoetica:
We've had the debate before about perma-removing single or double quotes from HTML. I favor the current standard of optional.

timhall:
I’m fine with the quotes, it’s the string vs tag issue that I was mainly addressing, where there are special cases where you can pass js expressions as a string instead of a tag (`{...}`). To make it more confusing, there are some cases where tags aren’t allowed (actions, I think).

Unfortunately, this leads to the quote/unquoted debate since `“{ ... }”` is best supported by editors, but is this a string with an tag inside? and, if so, should it be interpreted as the value’s type or a string? I’m sure this has been debated (and decided) previously, so I picked the unquoted form to try to express the string vs tags issue.

mrkishi:
> The above should satisfy the current store needs, although there is no built-in `$...` in templates, which is a bit of a shame (although arguably less magic). This could be take a step further and generalized to support component context and the idea of connecting to observables / subjects = `Context` (🐃).
> 
> ```ts
> interface Context<T> {
>   get(): T;
>   set(T): void;
>   subscribe(callback: (T) => void): () => void
> }
> ```

Nice, that's inline with the [repl](https://svelte.technology/repl?version=2.15.1&gist=606a52eaae3b41300a7ccbea4a08fe81) I created the other day.

ryanatkn:
For stores, I'll echo the above comments, that I like both reusing reactive assignment and having something generic like [React's context](https://reactjs.org/docs/context.html) that gives descendant components at any depth a way to access stuff (including stores) provided by an ancestor.

As a thought experiment, could a `store` just be a component? Like all components, they have internal reactive assignment semantics and an exported contract to the world. They could be invisible wrappers with just a `<slot>` or render whatever they want. They could be imported by any descendant via context, maybe with an API like [React's `useContext`](https://reactjs.org/docs/hooks-reference.html#usecontext). The idea of a `store` would become a convention in a system where descendant components can load any contexts provided by ancestors.

This seems weird and I don't know all of the tradeoffs and implications. I like the outcome of removing the standalone store concept while giving components more powers for composition. 

A possible API would have descendants directly import the ancestors they want. This seems simple and friendly to type systems, but implies that circular deps need to work.

```js
<script>
  import CornerStore from './CornerStore.html';

  let cornerStore; // type inferred as `Context<CornerStore>` (`| undefined`?)
  useContext(CornerStore, (c) => cornerStore = c);
</script>
```

So in React terms, the context object, the value returned from `React.createContext` that's used to declare which type of context you want, would just be the imported component, and components provide a context of their type to descendants, either implicitly or explicitly. I can't tell if that's an unfortunate restriction or another good unifying simplification.

Rich-Harris:
The success of these ideas hinges on there being as little compiler magic as possible. Currently, there are two bits of magic:

* assignments are instrumented with a call to `__update`
* `export` is removed from your code, as it is purely a compiler directive

The reaction we've had so far suggests that people are on board with both of those rules. But if we start adding more rules — such as changing the meaning/behaviour of specific function calls — we'll quickly find ourselves in the uncanny valley, and people will struggle to understand what is happening to their code.

That's why we need `createEventDispatcher` instead of just `dispatch` (or `fire` if you prefer). In this code...

```js
import { dispatch } from 'svelte';

setTimeout(() => {
  dispatch('hello', { name: 'world' });
}, 100);
```

...there is no way to know which component fired the event without doing something weird like this:

```diff
setTimeout(() => {
-  dispatch('hello', { name: 'world' });
+  dispatch('hello', { name: 'world' }, __self);
}, 100);
```

The same logic applies to `connect` and `context` — I think it becomes a lot harder to understand how data is flowing around the system. I think we're in a much better spot if we rely on JavaScript itself, even if it means typing slightly more:

```js
let friends;

const listener = someThirdPartyAPI.subscribe(id, value => {
  friends = value;
});

ondestroy(() => {
  listener.unsubscribe();
});
```

It's easy enough to wrap that with a helper function, if you need to use it in multiple components:

```js
import { ondestroy } from 'svelte';

function subscribe(id, callback) {
  const listener = someThirdPartyAPI.subscribe(id, callback);
  ondestroy(() => listener.unsubscribe());
}

// used like so
let friends;
subscribe(id, value => friends = value);
```

In fact we could implement an equivalent of 'context' in the same way:

```js
function createContext(start, value) {
  const callbacks = [];
  let stop;

  function set(newValue) {
    value = newValue;

    callbacks.forEach(callback => {
      callback(value);
    });
  }

  return {
    subscribe(callback) {
      if (callbacks.length === 0) {
        stop = start(set);
      }

      callbacks.push(callback);
      callback(value);

      ondestroy(() => {
        const index = callbacks.indexOf(callback);
        callbacks.splice(index, 1);

        if (callbacks.length === 0) {
          stop();
          stop = null;
        }
      });
    }
  }
}
```

```js
// src/contexts.js
import { createContext } from 'svelte';

export const online = createContext(
  function start(set) {
    function handler() {
      set(navigator.onLine);
    }

    window.addEventListener('online', handler);
    window.addEventListener('offline', handler);

    return function stop() {
      window.removeEventListener('online', handler);
      window.removeEventListener('offline', handler);
    };
  },
  navigator.onLine
);
```

```html
<script>
  import { onlineContext } from './contexts.js';

  let online;
  onlineContext.subscribe(value => online = value);
</script>
```

(I'm using the name 'context' here since it replicates the functionality of context, but really it's just a take on the concept of observables. And I'm not saying I think we should add this to Svelte, just throwing it out there as a possibility — it might be better as a non-Svelte-specific thing, i.e. without `ondestroy`.)

---

Some version of 'context' is definitely important — the difficulties with using multiple stores in a Svelte 2 app have shown that to be the case. It's made more urgent by the fact that you can't declaratively attach a store to an arbitrary component in this RFC; you can only attach a store at the top-level when you do `new App({..., store})`.

It would be nice to have unification between whatever concept of 'store' and 'context' we land on, but I do think there is still value in the concept of a single app-level store (particularly one that can easily be referenced in markup), not least because it's turned out to be such a useful way to pass data from server to client (such as user data) in Sapper.

---

I'm interested in discussing the thoughts around changing directive syntax a bit more. There possibly are some inconsistencies at present. 

That said my inclination is to avoid going too far with the curly braces — I'm not sure I see the advantage of this...

```html
{#if visible}
  <p in:{fly}={{ y: 50 }} out:{fade}>fades in and out</p>
{/if}
```

...over the status quo:

```html
{#if visible}
  <p in:fly={ y: 50 } out:fade>fades in and out</p>
{/if}
```

This is a little harder to explain but I think of the contents of `{...}` as being read-only views onto expressions that could contain any JS — there is a significant difference between `foo={bar}` (which we can think of as `foo <= bar`) and `bind:foo=bar` (which we can think of as `foo <=> bar`). In the first case we could replace `bar` with a different expression (`foo={42}`) and it makes total sense, but `bind:foo={42}` is in a way equivalent to `var 42 = foo` (which of course is invalid code).

Directives are probably the least well rationalised part of HTMLx though, so I'm open to changes that make them more consistent without harming read/writeability.

tomcon:
re. new Store & following on from previous discussion:
I'm with @tivac on this one.
The (relatively) simple & consistent (ie. same api) implementation of Store in v2 won a lot of people over imo - including me! By all means knock yourself out with other stricter store implementations etc but in some parts of the real world it is ***very*** useful indeed

arxpoetica:
I'm also w/ @tivac

evs-chris:
This looks pretty awesome! Coming from ractive, I feel like a good store is a pretty important thing, so I'm going to babble incoherently a bit.

I really like the above idea of implementing the store as a sort of adapter (`Store { get; set; subscribe; }`) with a default implementation provided if the component has requested store access (by referencing it) and none was provided by the user. It _is_ a bit magical, but it could be entirely opt-in magic in the form of a single well-defined macro:

```js
// compiler, I'd like a bit of that sweet sweet magic
import bind from 'svelte/store/bind';

// private state
let { something } = bind();

// public state via instance property
export let { other } = bind();

// I wouldn't hate keypath-style bindings, but I don't know how well that works here :)
```

The `bind` macro could then expand to create the local state that pulls from the store, subscribe for updates with the subscription managed by the component lifecycle, and set up participation in the `__update` cycle for the component with the references in the store. Precedent for slightly magical non-functions already exists in ES in the form of `import()`, and you could still wire things up manually, if you're magic averse. I believe you'd still need a way to inject a store reference into the component, which could just be an import too? `import store from 'svelte/store';` where behind the scenes that injects either the provided store or provides a default. Components could automatically have their owner's store, a store supplied via directive, or a default injected.

At this point, referencing a store variable in a template could create an implicit state `let`, like the implicit shorthand for regular state, if it's not already declared for API purposes. The `__update` and subscription mechanisms would also need to ensure that they don't cycle, which is another reason that some sugar that handles it correctly for you would maybe be nice.

kylecordes:
> solve the problem of reactivity at the language level

Even before considering any of the details, embracing this idea is a profound step in the right direction. This is what we need languages to do, rather than to need app code full of wiring to library level reactivity, to increase productivity.

> exported values form the component's contract with the outside world

I'm mentally comparing this with similar use of language features that some other frameworks take, in which you provide a class, and the public members of that class become the contract with the outside world.  Although classes are kind of out of fashion, this seems like a case where they are worth consideration, fitting the semantics at hand? But...

On the other hand, the class approach often ends up with an extra layers of language construct. Consider what Angular does: If you follow the Angular style guide, you always write just one component in a source file, which is a JavaScript module. So you have the JavaScript module layer acting mostly as boilerplate, with just the class (component) inside specifying an API for the outside world. The approach described in the RFC avoids this, tightening up the syntax and reducing boilerplate. Arguably there is no need for a class to describe a component, because a JavaScript module already is a somewhat class-like construct that describes a component.

> createEventDispatcher

What I don't like adore this approach so far, is that it still necessary to actually execute all possible code paths to find out what events a component might emit. Ideally the list (and shape) of emit-able events would be part of the declared public API of a component. The tooling (compiler) would produce reasonably complete metadata about the outside surface of a component, including the events.

How about an approach that elevates the tick and talk in the example, to become language identifiers that are exported somehow?

```
export const tickEvent = …
```

> Server-side rendering

Since computer effort is so much cheaper than human effort, perhaps consider scrapping the `generate` option? The compiler could always emit both, and always require that components be written such that they can successfully compile either way.

(I think the future of higher productivity development will be improved by raising the bar, building slightly less flexible tools in exchange for those tools "always" producing more consistent, coherent output. is it okay to say to Svelte users, "Svelte is intended to make 100% components that work seamlessly in the browser or SSR, if you don't want that, consider one of the 300 older, less mature component mechanisms"?)

> Sync vs async rendering

It is a bit sad to lose a point of differentiation, but this change is probably inevitable. I think other frameworks moving in a similar direction have already reached this conclusion as inevitable also.

> Svelte would continue to offer a custom element compile target

Along the same lines as earlier, how about if this was not a switch, but rather nailed on?

* Write a Svelte component.
* Compile it. The output includes all of:
  * the smallest possible Svelte-only component output
  * web component output
  * SSR output.

This would close off the avenue of "create hundreds of components without web component compatibility in mind, or SSR in mind, discover you want those, and realize there is a whole bunch of work to do”. (I don't know if this matches the Rich vision!) 

# Summary

Hopefully some of my rantings are helpful. Overall this RFC is excellent and I think a design anywhere in the general vicinity will be a great future for Svelte!

arxpoetica:
@PaulMaly noooo! They be stealing my two-way bindings! I love the power of two-way bindings. Simple and elegantly solves some very important things without the boilerplate of events.

As for the following, I love this proposal:

```
import { on } from 'svelte';
on('destroy', () => { ... });
```

Kiho:
I like 2 way binding too, on:input, on:update .... does not work well with dynamic component.

PaulMaly:
Hi @arxpoetica @Kiho!

> They be stealing my two-way bindings! I love the power of two-way bindings. Simple and elegantly solves some very important things without the boilerplate of events.

> I like 2 way binding too, on:input, on:update .... does not work well with dynamic component.

Actually, I agree with you guys! But in RFC we read:

> The reason we can consider removing component bindings is that it's now trivial to pass state-altering functions down to components:

I don't like this idea and for me tags 2way bindings less useful than component bindings. I think if we're removing component bindings, 2way loses meaning.

> If we just use 3rd party store then I don't think we can have context based sub-store.

Could you please describe a little detailed what do you mean? In this part, I've guided a principle: "Svelte provide only that features which can't be implemented in userland". Seems v3 gives us more opportunities to use 3rd-party stores. So, I believe we could discuss are we really need own solution.

PaulMaly:
Improve my initial comment a little bit:

1. We need only a few helpers functions in svelte. Each function should automatically be associated with the component instance during instantiation:

```javascript
import { destroy, dispatch, on } from 'svelte';
```
These helpers should be attached as public interface of component too.

2. I think good idea to use default export for meta info and compiler settings of component.

```javascript
export default {
  tag: 'my-component'
};
```

3. I believe we should leave 2way bindings for component or remove this feature completely.

4. Multiple `<script>` tags for differences environments is a good idea for Svelte/Sapper, in my opinion. Seems, some kind of operations works really differently on that environments with almost the same result. This is very clean way to describe it.

5. Event listeners callback using `on` method should receive apropriate `event.detail` object with event-specific stuff.

```javascript
import { on, dispatch } from 'svelte';

on('props', e => {
    console.log(e.detail) // all component props
});

on('update', e => {
    console.log(e.detail) // all component state
});

dispatch('customEvent', { payload: 'something' });

on('customEvent', e => {
    console.log(e.detail.payload);
});

```

6. I believe it'll be nice if Svelte become more lazy and async. Maybe more compiler-driven memoization things. I prefer if we move to async rendering and all that things like time-slicing etc.

7. Seems, we could rethink do we really need to have built-in Store solution?

PaulMaly:
One more insane idea, perhaps, component's instance could be represented with real DOM element extended by exported props and other stuff. Seems, the closest analogy is Shadow Root. 

So, we'll be able to all power of Svelte feature and DOM API as well:

```javascript
import Component from '...';

const comp = new Component({
    /* props here */
});

comp.addEventListener('update', () => {...});

comp.foo = 'update component prop';

document.body.appendChild(comp);
```

In this case, `dispatch` & `on` methods would be unnecessary or become just a sugar:

```javascript
comp.addEventListener('customevent', () => {...});

comp.dispatchEvent(new CustomEvent('customevent', { detail }));
```
We also will be able to use MutationObserver for the components directly.

Rich-Harris:
Thanks everyone! Big ol' response dump inbound:

> a single well-defined macro

I'm *really* keen to avoid these :) At the very least, I think they should be a last resort. *So far* I'm not persuaded that we need them, though I realise we haven't come up with a really compelling solution for a v3 Store.

> it still necessary to actually execute all possible code paths to find out what events a component might emit. Ideally the list (and shape) of emit-able events would be part of the declared public API of a component...How about an approach that elevates the tick and talk in the example, to become language identifiers that are exported somehow?

I do like the idea of events being part of a declared API — would be great to (for example) get an edit-time warning that `on:select` should actually be `on:selected`, or whatever. Having said that, it becomes more complex if events bubble, since the component you're adding the listener to may not be the component that emits the event. 

For the majority of cases you *could* determine which events a component was capable of firing, just by looking at the first argument of calls to `dispatch`. Not bulletproof (especially if `dispatch` was passed outside the component), but enough to provide information for your linter etc in most situations.

> Since computer effort is so much cheaper than human effort, perhaps consider scrapping the `generate` option? The compiler could always emit both, and always require that components be written such that they can successfully compile either way.

I love this idea! We could add a static `$render` named export to components, mirroring the `$on` and `$destroy` instance methods. This makes a ton of sense. We'll have to be even more careful than we already are about maintaining tree-shakeability, but aside from the fairly modest overhead of generating some extra JS for bundlers to ignore, this sounds like all win.

I'm not sure if we can do the same thing for web components though. Typically they way you'd use a WC is with an empty import...

```js
import './my-app.html';
document.body.innerHTML = `<my-app></my-app>`;
```

...which implies that the component is self-registering, defeating tree-shakeability. It could export a `$define` or `$register` function instead, but that becomes slightly more cumbersome to consume.

There's also ideas for alternative compile targets (like [svelte-gl](https://github.com/Rich-Harris/svelte-gl)) kicking around, and we definitely wouldn't want to bundle all those in, so I think it probably does make sense to maintain these as a compiler option.

> In all examples <script> tag in top of the component. Is this will be mandatory or just for good looking?

Not mandatory. I've adopted that convention mostly because it's unusual to see a name used before its declaration (particulary with `let` and `const`, where it's a TDZ violation) — it feels more natural to me to have the declarations upfront. But it definitely won't be enforced by the compiler.

> Should we really have a differences between component's custom events and component's "lifecycle events"?

`on` as a standalone function makes much less sense *within* a component now that there's no `this` — there's basically no situation where you want a component to listen to an event that is emitted by the *same* component. Additionally, having separate lifecycle functions is good for things like tree-shakeability — if `on('update', ...)` is a possibility you need to include the `update` lifecycle stuff even if no-one is actually using it.

> How about use default export for this purpose? In v2 all our default export looks like compilator settings.

In v2, you have to write that `export default` in a very specific way. You can't, for example, do this:

```js
import { tag_prefix } from './config.js';

export default {
  tag: `${tag_prefix}-thingy
};
```

In v3 we can completely jettison basically all the confusing limitations about what (perfectly valid!) JS doesn't work for the compiler. Preserving compiler options on `export default` would be therefore be a bit anomalous. 

> I think if we'll use vanilla CustomEvent, what if we don't need to have some Svelte-specific apis? Do we really need to have built-in event system?

This is an interesting idea. Definitely warrants further consideration.

> I belive we should refuse 2-way bindings

I'd argue `<input bind:value>` is sufficiently nicer than `<input value={val} on:input={e => val = e.target.value}>` that it deserves to survive v2-v3. With components there's an extra consideration — in this example...

```html
<Component foo={foo} on:update={e => foo = e.detail.foo} />
```

...if the child changes the value of `foo`, the parent re-renders, which involves *passing foo back to the child*. A `bind` directive gives the compiler the opportunity to track which *direction* the data is flowing in, preventing unnecessary work or (in the worst case) an infinite update loop. (This does get harder in v3 because everything is async, it's worth noting.)

> If we'll able to have multiple <script> tags in components, I think could albe to have separated scripts for client and server

I think there's potential for this idea, but I also think that *in general* we should strive to make the behaviour of components as consistent as possible between client and server, and the framework should encourage that. Adding extra script blocks is likely to result in less readable and less DRY code.

Of course, it's something that could be implemented in a `preprocess` step fairly easily, so we could experiment with this idea in userland before committing to adding something to the compiler itself.

> I think we can provide all props in event's detail object and use it like this

Definitely a possibility. I kinda like the idea of `onprops` not having an argument, because the idea is that the `foo` in `export let foo` has already been updated in scope by the time `onprops` runs (you don't have to do `foo = newProps.foo` or anything), and because it gives us more flexibility to change our minds in future. But it's not set in stone.

> `<div>{store.foo}</div>`

The trouble there is that there's no (good) way to make that expression *reactive*. I have been wondering if that's what the job of `$` should be (i.e. `$store.foo` says '`store` is an object that implements the Store API, and we want to use that API to observe `foo`'. But I don't know.)

> One more insane idea, perhaps, component's instance could be represented by real DOM element extended by exported props and other stuff.

That's not insane, it's called [Custom Elements](https://svelte.technology/guide#custom-elements) and it's already implemented :)

arxpoetica:
@Rich-Harris:
> there's basically no situation where you want a component to listen to an event that is emitted by the same component

Come again? What does this mean?

Rich-Harris:
Events are for knowing what's happening with *other* objects. There's no value in this:

```js
component.on('message', message => {
  console.log(message);
});

component.fire('message', 'hello, err... me');
```

arxpoetica:
> $store.foo says 'store is an object that implements the Store API, and we want to use that API to observe foo'.

I like that! Or maybe just `$.foo` but that starts looking an awful lot like another familiar API...

PaulMaly:
@Rich-Harris, actually, I'm not sure we understand each other correctly. To prevent any misunderstandings, I try to describe controversial points more deeply:

> there's basically no situation where you want a component to listen to an event that is emitted by the same component.

I'd argue, for example, I need to make some side-effect when some prop or internal state values are changed. I we'll be able to use `onprops` & `onupdate` callback multiple times, like this:

```javascript
onprops(() => {
...
});
....

onprops(() => {
...
});
```
There's not any difference between that and events usage. But events win in this case because we automatically have an opportunity to listen to them outside of the component. 

> Events are for knowing what's happening with other objects. There's no value in this:

Events are that the source will want that they were. Event producer able to put any payload to the event. So, seems to me, that callback become redundant in the case when we anyway need to have external events corresponding with `onprops` & `onupdate` callbacks.

> Additionally, having separate lifecycle functions is good for things like tree-shakeability — if on('update', ...) is a possibility you need to include the update lifecycle stuff even if no-one is actually using it.

I'm not sure about it. Seems, the compiler can't be absolutely sure that external event `update` and `props` not using somewhere outside of the component.

```javascript
const comp = new Component({ ... });
...
comp.on('update', () => {
    /* external code wants to know when an internal state changed */
})
```
So, you need to include this lifecycle stuff anyway. I believe it's not a problem because component needs to have minimal and predictable external api.

> In v2, you have to write that export default in a very specific way. You can't, for example, do this:

Actually, this case is really unusual. I don't think that we could to think up many such cases. I'm not sure I will ever want to have a dynamic name of the tag. On the other side really ugly `<svelte:meta>` solution. Why compiler options should be described in my templates? In html? Btw, for the external observer, it looks more like creating a regular html meta-tags:

```html
<svelte:meta> 
<!-- should become ? -->
<meta ... >
```
I believe compiler options could be and maybe even should be static.

```html
<script>

/* very clean and habitually */
export default {
  immutable: true,
  tag: 'my-tag'
};
</script>

<!-- really ugly -->
<svelte:meta immutalbe={true} tag="my-tag">
```

Besides, if we'll ever need to have some static function for configuration or any other purpose, I'll be able to use it with `export default` in a habitual way:

```javascript
export default {
  preload() {
     /* hi sapper */
  }
};
```

> I'd argue <input bind:value> is sufficiently nicer than <input value={val} on:input={e => val = e.target.value}> 

I agree, but my point is - if we'll consider removing component bindings, I think element's bindings are also should be removed, because it have much less meaning than component's bindings.

> Definitely a possibility. I kinda like the idea of onprops not having an argument, 

I believe this is the most clean solution for spreading props.

> The trouble there is that there's no (good) way to make that expression reactive.

Perhaps, we don't need to make that expression reactive? Seems compiler will not be able to make reactive all imports, right?

```html
<script>
  import { some_object, some_variable } from '3rd_party_lib';

  let foo = 0; // 'this is internal state and will be reactive'
</script>
<div>
   <button on:click={e => foo += 1}>Increment</button>

   <!-- how we can make this reactive? -->
   <button on:click={e => some_variable += 1}>Increment</button>
</div>
```
My point is, in v3 we'll not able to make all imports are reactive, but we'll give an access to any import in our templates. So, maybe we can just use api of these external modules as is?

```html
<script>
  import { some_object, some_variable } from '3rd_party_lib';

  let foo = 0; // 'this is internal state and will be reactive'
</script>
<div>
   <button on:click={e => foo += 1}>Increment</button>

   <!-- we can just use an api of external libs ? -->
   <button on:click={e => some_variable.set(some_variable + 1)}>Increment</button>
</div>
```

> That's not insane, it's called Custom Elements and it's already implemented :)

Not really :) It's more looks like Custom Elements emulation. I'm talking about when `customElement: false` in compiler. 

Right now Svelte component is a constructor which returns an object, but how about to return actual DOM node with extended api? This node can be considered by us as a certain analog of shadowRoot. We can wrap all our component's templates to one additional node, extend that node with Svelte-specific things and return it as imperative version of component for external usage. This approach will gives us full power of DOM API right over svelte component and also will increase compliance with the standard Custom Elements where components has shadowRoot node.

```javascript
const comp = new Component({ ... }); 

comp.querySelector('.some');
comp.addEventListener('click', () => {});

document.body.appendChild(comp);
```
You can check how it works in Angular [here](https://toddmotto.com/emulated-native-shadow-dom-angular-2-view-encapsulation).

mindrones:
typo: `<input on:keydown"{e => e.which === 39` (missing "=")

text/0001-reactive-assignments.md
```diff
@@ -0,0 +1,1068 @@
...
+There are three other lifecycle functions required — `onMount` (similar to `oncreate` in Svelte v2), `beforeUpdate` (similar to `onstate`) and `afterUpdate` (similar to `onupdate`):
+
+```html
+<script>
+  import { beforeUpdate, afterUpdate, onMount } from 'svelte';
+
+  export let foo;
+
+  beforeUpdate(() => {
+    // this callback runs whenever props change
```
timhall: I believe this also runs when reactive assignments are made, so I think "props" here may be confusing since I understand props to only be those values that are exported.
```suggestion
    // this callback runs whenever props or reactive variables change
```

text/0001-reactive-assignments.md
```diff
@@ -0,0 +1,1068 @@
...
+
+We toyed with a few alternative concepts:
+
+* Not doing anything, and attempting to fix the niggly edge cases within the current paradigm
+* Adopting something akin to React Hooks. This was met with a negative reaction from the community. Despite some minor ergonomic advantages in certain cases, the downsides of Hooks (the 'rules', the reliance on repeatedly calling user code, etc) were considered greater than the advantages
+* Pursuing a purer vision of reactive programming, akin to [that described by Paul Stovell](http://paulstovell.com/blog/reactive-programming). This would arguably make code more difficult to reason about, not less, and would likely introduce difficult syntactical requirements
+
+## Unresolved questions
+
+The major TODO is how `Store` should evolve in this new world.
```
timhall: I believe this is resolved now by #5

text/0001-reactive-assignments.md
```diff
@@ -0,0 +1,1068 @@
...
+### Custom elements
+
+Svelte would continue to offer a custom element compile target. No real changes would be involved here, since this proposal brings 'vanilla' Svelte components in line with web components viz. property access and event handling.
+
+The `props` compiler option is redundant now that the contract is defined via `export`ed variables. The `tag` option can be set via `<svelte:meta>` as discussed above.
+
+
+### Dependency tracking
+
+In Svelte 2, 'computed properties' use compile-time dependency tracking to derive values from state, avoiding recomputation when dependencies haven't changed. These computed properties are 'push-based' rather than 'pull-based', which can result in unnecessary work:
```
timhall: Is this section still valid with the reference implementation of #8? I think #8 is a much cleaner solution, I wonder what would be the expected behavior of function calls in html. I think the most sensible approach would be to re-run the function only when reactive arguments change.

```html
<script>
import { fn } from './functions';
</script>

<ul>
  <li>{fn()} - static, called on initial render</li>
  <li>{fn(a, b)} - dynamic, called whenever a or b change</li>
</ul>
```

nolanlawson:
Thanks for writing this up. I haven't read all the comments, but I read the RFC, and I have a few concerns:

1. Do we have evidence that the new system will result in smaller/faster code? I see that the method names and property names can be minified now (not that they couldn't before: https://github.com/sveltejs/svelte/issues/1102#issuecomment-382507126), but other than that I'm not sure if anything will actually get faster from the end-user's perspective as a result of these changes.
2. Linters will warn on unused variables in this new system. Maybe not a huge deal, but breaking that contract either means I need to carefully apply `eslint-disable` comments or forgo the benefits of unused variable detection entirely.
3. In the new system, it seems a lot more difficult for the developer to understand where computed property costs are coming from. In the old system, I could glance at the `computed` object and get a rough idea of the dependency tree, and see "Aha, this computation is expensive, so I probably want to make sure that the inputs aren't Objects or Arrays so that we don't re-compute unnecessarily." In the new system, using just functions, it seems a lot more difficult to guess what the compiler is doing under the hood because we don't have a straightforward list of inputs/output.
4. If I want to measure an expensive `set()` call in Svelte 2 (e.g. using `performance.mark()` and `.measure()`), I just wrap the `set()` call in my `mark()`s. In this new system, would this work the way I think it would?

```js
let foo
function recompute() {
  performance.mark('start')
  foo = doExpensiveFooCalculation()
  performance.measure('total', 'start')
}
```

Again, some of this is hard to predict because of the compiler magic.

Overall, I think this proposal is very ambitious and interesting, and it's definitely going to push Svelte in a wild new direction. I'm a little wary of introducing such a huge new change, though, as it will take a long time to upgrade a large app like Pinafore. Also, I for one actually enjoy the `data`, `store`, and `computed` system.

Conduitry:
Re: the linting issue, yeah this change does make it less useful to just use e.g. `eslint-plugin-html` - but I've been working on a Svelte 3 ESLint plugin - https://github.com/Conduitry/eslint-plugin-svelte3 - and I'm pretty happy so far with the experience it's able to provide.

Rich-Harris:
> 1. Do we have evidence that the new system will result in smaller/faster code?

This is tricky to quantify without actually converting an app. The smallest apps will get larger, because there's a bit more in the way of scheduling code etc, but my gut tells me that larger apps will get smaller, for various reasons (more things can be hoisted out of `ctx`, Svelte can more easily determine which things don't need to have update code generated for them, etc). I'm hoping to try it out this weekend on a couple of apps (just need to update Sapper...) and will report back.

In terms of performance, it's not so much that it'll become faster or slower across the board, but rather that performance will be significantly improved in the *worst* cases. In Svelte 2 it was much too easy to end up in a situation where you were doing interleaved DOM reads and writes, because of the synchronous updates — this is no longer the case. Longer term we should be able to finesse this further (for example being smart about when transitions are triggered, causing DOM measurements). 

> 2. Linters will warn on unused variables in this new system

It's true that there'll be some finagling involved depending on your setup. As @Conduitry says this *is* solvable. In Svelte 2 I always have a problem that *isn't* solvable AFAICT — referring to built-in-methods inside hooks and custom methods causes red squiggles, because VSCode is smart about figuring out object shapes etc, but (justifiably) not smart enough to know the idiosyncracies of a Svelte `export default`. So we're effectively exchanging a migraine for a mild headache.

> 3. In the new system, it seems a lot more difficult for the developer to understand where computed property costs are coming from

Agree. I think I need to update this RFC with the latest thinking — basically, I came to the conclusion that tracking which values are read during a function invocation isn't going to work, and would result in unpredictable performance characteristics. Instead, we're pursuing an idea that's more similar to Svelte 2's `computed` in [RFC 3](https://github.com/sveltejs/rfcs/blob/reactive-declarations/text/0003-reactive-declarations.md). It's a slightly radical idea, but so far it's proven really nice to work with.

> 4. If I want to measure an expensive `set()` call

There are two parts to this — the `doExpensiveFooCalculation()` and the resulting update. Svelte 3 doesn't *transform* your code, so the `performance.mark`/`performance.measure` will work exactly as you'd expect. It just instruments it with a call to an internal function that invalidates `foo` and schedules an update (in a microtask) if one isn't currently scheduled.

The update itself is in the framework's control. There are a couple of ways you could profile that — by putting `performance.mark` in `beforeUpdate` and `performance.measure` in `afterUpdate`, or by manually calling `flush` (which is currently exported from `svelte/internal.js`. If it proves necessary, `flush` could be re-exported from `svelte` itself).

---

I definitely recognise that this is going to be a pain to upgrade, and I regret that. I do think it's necessary though — a [lot of issues](https://github.com/sveltejs/svelte/labels/rfc-1) are extremely hard to solve in Svelte 2 but will become easy (or irrelevant) in Svelte 3, and it allows us to jettison a lot of compiler complexity which will let us move faster. There'll be a lot less for newcomers to learn, and components will be easier to write.

Do keep the feedback coming, particularly if there are elements of the 3 RFCs that haven't been explained as well as they could be.

Conduitry:
Perhaps we should remove the bit about function call dependency tracking, link to the reactive declaration RFC, and merge this, as it's fairly settled now.

Rich-Harris:
Agree re removal — have updated it. I think we're supposed to announce a 'final comment period' before merging, if we're respecting the RFC process... I guess we should do that if the three v3 RFCs are in good shape

PR #6, opened 2018-11-23
CSS-in-HTML

arxpoetica:
**[Rendered](https://github.com/arxpoetica/rfcs/blob/css-in-html/text/0000-css-in-html.md)**

Even though this is an incomplete proposal, there's been so much conversation in the #future Discord channel, I wanted to give us a place to record some of the conversation more inline with the (ongoing) proposal.

Expect some of those conversations to start to land.

If we find this PR to become too encumbered with old ideas, I'll close it and reopen a new one after incorporating some of the top / sanctioned ideas.

But for now, consider this just a good place to capture ongoing discussion.

### DIFF
--- a/text/0000-css-in-html.md
+++ b/text/0000-css-in-html.md
@@ -1,0 +1,112 @@
+- Start Date: 2018-11-05
+- RFC PR:
+- Svelte Issue:
+
+### THINGS TO WATCH:
+
+https://www.youtube.com/watch?v=lK3OiJvwgSc
+https://github.com/GoogleChromeLabs/houdini-samples/blob/master/layout-worklet/masonry/index.html
+http://ishoudinireadyyet.com/
+https://adamwathan.me/css-utility-classes-and-separation-of-concerns/
+
+
+### WIP from the mouth of Rich:
+
+* modularity. we need a way to import CSS that the compiler understands, and we need a way to compose styles a la modular-css
+* optimisation. at present, the only form of optimisation is (incomplete) minification
+* as you put the RFC together for the dynamic parts, I'd urge you to think about implementation approaches and how they affect the design. For example, if I have a class like .foo and some markup like <div class={passedInFromOutside}> the compiler doesn't know that class="foo" could be the result, so it becomes incredibly hard to know that it needs to apply particular dynamic styles to that element
+* (if the implementation calls for CSS variables, then we should probably just use CSS variables in dynamic style tags, and be explicit about the whole thing)
+* I'd also suggest studying how theming is applied to custom elements with shadow DOM (I'm not sure what the current state of the art is, I only know that shadow-piercing selectors were deprecated). This is for two reasons: 1) they will have already discussed and anticipated a lot of the problems we would likely face, and 2) it makes it more likely that we're able to reuse the solution when compiling to custom elements(edited)
+* For the optimisation piece, it's worth looking at OptiCSS and thinking about that work could apply to ours (Svelte could even use OptiCSS directly, perhaps). That will likely have ramifications for the compiler API (since it needs to think beyond single components), so it touches on the question of whole-app optimisation
+* CSS Shadow Parts https://drafts.csswg.org/css-shadow-parts/
+
+# CSS-in-HTML
+
+## Summary
+
+This is a larger RFC focused on solving CSS scoping on both a stand-alone component and global level. This RFC also proposes an approach to reactive CSS as well as better tooling for incorporating 3rd party processors and compilers.
+
+## Motivation
+
+Svelte developers generally acknowledge that there needs to be a better answer for CSS both in components as well as a larger ecosystem consideration. There are varities of opinions about best methodologies for CSS. This proposal sidesteps opinions as much as possible, allowing developers to work with whatever framework or method best suits their purposes, yet cherrypicking the best design considerations.
+
+One debate often boils down to scoping considerations. Currently, with Svelte 2, the inclusion of any CSS within a style tag is considered 100% scoped. Consider:
+
+```html
+<!-- main component -->
+<div>
+	<CascadingDiv/>
+</div>
+<style>
+	div { color: goldenrod; }
+</style>
+<script>
+export default {
+	components: { CascadingDiv: './cascasing-div.html' }
+};
+</script>
+```
+
+```html
+<!-- component: cascasing-div.html <CascadingDiv/> -->
+<div>This is not goldenrod. Styles do not cascade!</div>
+```
+
+The only way out of this is to wrap your selector(s) in `:global()`. This, however, gets to be fairly verbose when you want to wrap all selectors on the page.
+
+```html
+<style>
+:global(.foo) { text-align: center; }
+:global(.foo span) { font-size: 80%; }
+:global(.foo em) { color: pink; }
+.bar { text-align: right; }
+<!-- unscoped: `.foo *` -->
+<!-- scoped: `.bar` -->
+</style>
+```
+
+Arguably, the appeal of single file components is encapsulating all component-relatives into the single file. But the above also has the sometimes unwanted effect of cascading everywhere. This presents the complex challenge of controlling scope to a single component, a set of embedded components, or global.
+
+It would be nice if there was a simple mechanism for controlling scope more easily at a component level, making JS variables behave dynamically in CSS and HTML, as well as passing style scopes from component to embedded component if necessary.
+
+## Detailed design
+
+> This is the bulk of the RFC.
+
+> Explain the design in enough detail for somebody
+familiar with the framework to understand, and for somebody familiar with the
+implementation to implement. This should get into specifics and corner-cases,
+and include examples of how the feature is used. Any new terminology should be
+defined here.
+
+## How we teach this
+
+> What names and terminology work best for these concepts and why? How is this
+idea best presented? As a continuation of existing Svelte patterns, or as a
+wholly new one?
+
+> Would the acceptance of this proposal mean the Svelte guides must be
+re-organized or altered? Does it change how Svelte is taught to new users
+at any level?
+
+> How should this feature be introduced and taught to existing Svelte
+users?
+
+## Drawbacks
+
+> Why should we *not* do this? Please consider the impact on teaching Svelte,
+on the integration of this feature with other existing and planned features,
+on the impact of the API churn on existing apps, etc.
+
+> There are tradeoffs to choosing any path, please attempt to identify them here.
+
+## Alternatives
+
+> What other designs have been considered? What is the impact of not doing this?
+
+> This section could also include prior art, that is, how other frameworks in the same domain have solved this problem.
+
+## Unresolved questions
+
+> Optional, but suggested for first drafts. What parts of the design are still
+TBD?


arxpoetica:
Css vars via @lukeed:

```
<div class="demo">    
    <button on:click={() => value++}>Increment</button>
    <button on:click={() => value--}>Decrement</button>
</div>

<script>
    export let value = 0;
</script>

<style>
    .demo {    
        padding: 10px;
        border-radius: 4px;
        -webkit-appearance: none;
        font-style: italic;
        /* this is dynamic */
        color: {value % 2 ? green : red};
    }
</style>

<!-- 
  Output (legacy:true)
    needs :yak: symbol because 
    it writes inline style properties
    but will overwrite anything meant to cascade
-->

<style>
    .demo {    
        padding: 10px;
        border-radius: 4px;
        -webkit-appearance: none;
        font-style: italic;
    }
</style>

<div class="demo" style="color:green;">...</div>
<div class="demo" style="color:red;">...</div>

<!-- 
  Output (legacy:false)
    sets (--var) values inline, which
    will respect cascade bcuz does not 
    change the style-rules or their order
-->

<style>
    .demo {    
        padding: 10px;
        border-radius: 4px;
        -webkit-appearance: none;
        font-style: italic;
        color: var(--color);
</style>

<div class="demo" style="--color:green;">...</div>
<div class="demo" style="--color:red;">...</div>
```

arxpoetica:
A couple others for quick reference:

* https://github.com/sveltejs/svelte/issues/758
* https://github.com/sveltejs/svelte/issues/901

timhall:
### Global

This has been discussed elsewhere, but I think building off of #4 with the multiple script tag approach (with `context="..."`) could work well here:

- By default, all `<style>` tags are `scoped` to the component and has the current svelte behavior
- `<style context="global">` is used to scope all contained styles globally

```html
<style context="global">
  body { background: #111; }
  h1 { font-size: 3rem; }

  /* Equivalent to:
  :global(body) { background: #111; }
  :global(h1) { font-size: 3rem }
  */
</style>
<style>
  a { color: white; } 
</style>
```

### Cascade

For cascading styles (apply from component to children), `.container :global(...)` would likely still be needed since svelte doesn't have an explicit top-level element to attach styles to (i.e. need to define `.container` to attach cascading styles). I'd argue that for applying styles to child components, the child should have control over how those styles are applied. This could be through:

1. Properties - Component exposes semantic properties that map internally to styles
2. Classes - Component uses specific classes to target parts of the component
3. [CSS Shadow Parts proposal](https://drafts.csswg.org/css-shadow-parts/) - Use `part="name"` in html and `Component::part(name)` in css to target certain elements in the component

```html
<style>
  section :global(.Post__title) { color: blue; }
  section Post::part(body) { line-height: 1.5; }

  /* Equivalent to
  section.svelte-id1 .Post__title { color: blue; }
  section.svelte-id1 .svelte-part-body.svelte-id2 { line-height: 1.5; }
  */
</style>

<section>
  <Post primary title="..." body="..." />
</section>
```

```html
<!-- Post.html -->
<h1 class="Post__title">>{title}</h1>
<!--       ^ 2. Classes -->

<div part="body" style="color: {primary ? 'blue' : 'grey'}">{body}</div>
<!-- ^ 3. Parts                 ^ 1. Properties -->
```

lukeed:
RE @timhall's last point:

* Not sure how I feel about introducing a new `part` attribute when we already have named slots

* Nothing inside `Post` component indicated that `.Post__title` would _not_ be scoped. How would we know (from the Parents' perspective) that that class name was a valid destination? 

* I think if we are to use the Component's name directly within style selectors, it might make more sense to change that syntax a bit:

    ```html
    <style>
      section :global(...) { 
        /* something global */
      }
      section :Post(.Post__title) {
        /* stemming from Post's ".Post__title" class */
        /* I will work regardless of scoped/global output */
      }
      section :Post('slotname') {
        /* I am targeting the <slot name="slotname"/> element  */
        /* ...Or the "part"'s name, if we could not work with existing <slot> mechanics   */
      }
    </style>
    ```

    The idea here is that we follow the `:global` and `:local` convention (which are context/scoping flags, essentially) and using them to indicate a scope/context of the Component (`:Component`).

arxpoetica:
@timhall regarding scoping via separate `<style>` tags, that idea presents problems w/ preprocessors when one doesn't want to duplicate nested styles for the sake of separate tags. Illustrated example for clarity:

Method with scoped tags:

```html
<style context="global">
	.component {
		font: inherit;
		h1 { margin: 0 0 0.2rem; }
		h2 { margin: 0; }
		p { color: goldenrod; }
	}
	/* precompiles to:
		.component { font: inherit; }
		.component h1 { margin: 0 0 0.2rem; }
		.component h2 { margin: 0; }
		.component p { color: goldenrod; }
	*/
</style>
<style>
	/* overriding locally */
	.component {
		color: pink;
		font: 1.6rem/1.2 Helvetica;
		h1 { margin: 0; }
		h2 { margin: 0 0 2rem; }
		p { color: inherit; }
	}

	}
	/* precompiles to:
		.component {
			color: pink;
			font: 1.6rem/1.2 Helvetica;
		}
		.component h1 { margin: 0; }
		.component h2 { margin: 0 0 2rem; }
		.component p { color: inherit; }
	*/
</style>
```

Alternate proposal with scoped `@global` (🐃) selector:

```html
<style>
	.component {
		color: pink;
		font: 1.6rem/1.2 Helvetica;
		h1 { margin: 0; }
		h2 { margin: 0 0 2rem; }
		p { color: inherit; }
		@global {
			font: inherit;
			h1 { margin: 0 0 0.2rem; }
			h2 { margin: 0; }
			p { color: goldenrod; }
		}
	}
	/* precompiles to:
		.component {
			color: pink;
			font: 1.6rem/1.2 Helvetica;
		}
		.component h1 { margin: 0; }
		.component h2 { margin: 0 0 2rem; }
		.component p { color: inherit; }
		@global {
			.component { font: inherit; }
			.component h1 { margin: 0 0 0.2rem; }
			.component h2 { margin: 0; }
			.component p { color: goldenrod; }
		}
	*/
</style>
```

It's a pedantic difference, but I find the latter grouping / clustering much more easy to deal w/ arrange, and (in part) keep DRY.

arxpoetica:
@lukeed > The idea here is that we follow the :global and :local convention

I am in favor of (if not abandoning) those microscopic selectors (they cause a lot of clutter), at least replacing with something that encapsulates better, i.e., a (🐃) `@global` and `@cascade` selector, or something better. Maybe we don't get rid of `:global` entirely since there are probably good one-off use cases, but there's still the problem of repeating `:global` multiple times when enclosure is better.

lukeed:
Was just following the rest of V3's approach of reusing existing syntax. I'm not familiar enough with latest improvements to CSS modules/PostCSS, but to me it's more appealing to construct CSS in a way that these tools can understand, in the event I'm porting styles to & from Svelte SFCs to standalone stylesheets.

timhall:
@arxpoetica I think your example is a good case for preprocess plus [postcss-nested](https://github.com/postcss/postcss-nested) and [postcss-global-nested](https://github.com/a-x-/postcss-global-nested) and why building on the "standard"ish `:global` is good since we can leverage existing tools (which is the svelte way).

```css
.component {
  color: pink;
  font: 1.6rem/1.2 Helvetica;
  h1 { margin: 0; }
  h2 { margin: 0 0 2rem; }
  p { color: inherit; }
  :global {
    font: inherit;
    h1 { margin: 0 0 0.2rem; }
    h2 { margin: 0; }
    p { color: goldenrod; }
  }
}

/* Transformed in preprocess using postcss-nested and postcss-global-nested to */

.component {
  color: pink;
  font: 1.6rem/1.2 Helvetica;
}
.component h1 { margin: 0; }
.component h2 { margin: 0 0 2rem; }
.component p { color: inherit; }
:global(.component) {
    font: inherit;
}
.component :global(h1) { margin: 0 0 0.2rem; }
.component :global(h2) { margin: 0; }
.component :global(p) { color: goldenrod; }
```

Alternatively, I've run into many issues with using global css frameworks like bootstrap or tailwindcss and getting about a thousand unused style warnings so it'd be nice to have an escape hatch for processing style blocks from a global context.

---

For overloading `slot`, that is an interesting possibility, but I can't think of when you'd want to apply styles directly to the slot as opposed to the containing element. Also, you may not want the consumer overriding what's in the slot, just want to give a place to attach styles.

```html
<!-- Button.html -->
<button part="button"><slot name="button" /></button>
``` 

```html
<Button>Howdy!</Button>

<style>
Button::part(button) {
  /* Applies to the button element itself */
}
:Button('button') {
  /* Applies to what is contained in the slot */ 
}
</style>
```

timhall:
That's a very good point and I think a confusing part of the proposal, with `part="value"` using a string and `::part(value)` using a bare identifier. I think I prefer using a string in both places for consistency, will have to check if there are any established cases with pseudo selectors leaning one way or the other

lukeed:
**_Aside_**

FWIW, before continuing, I should clarify my idea about using `<slot>`s...

For it to be useful, Svelte slots would have to operate like Vue's in that they can exist on their own (Svelte's current & only approach) ***or*** be applied onto an another element (much like @timhall's `part=""` idea:

```html
<slot />
<slot name="foo" />
<div class="hello" slot="bar" />
```

> If this were possible, then `part=""` would be redundant & should not be added IMO.

However, even _if_ @Rich-Harris allowed this to happen (I don't think he's a fan? hence current implementation), I don't know if they realistically would be very enjoyable to use because you (as the Parent component) would need to know if the `slot` you're selecting is one that disappears or if it's named DOM Element.

I mean, either thru `<slot> | slot=""` or `part=""` you (the Parent) always require working knowledge of the Child components you're manipulating

🤔 

---

I'd be happy with this & think it solves a lot of the conversation re: Component style inheritance & overwrites:

```html
<!-- Actions.html -->
<div class="actions">
  <div slot="primary" />
  <div slot="secondary" />
  <button>Cancel</button>
</div>
```

```html
<!-- Parent.html -->
<Actions>
  <button slot="primary">Submit</button>
  <button slot="secondary">Edit</button>
</Actions>

<style>
:Actions(button) {
  /* apply to all <button> within Actions */
  /* ==> button x3 (Submit, Edit, Cancel) */
}

:Actions('primary') {
  /* applies to the "div[slot=primary]" element directly */ 
  /* ==> <div> x1 */
}

:Actions('primary') button {
  /* applies to the "div[slot=primary] button" element directly */ 
  /* ==> <button> x1 (Submit) */
}

:Actions('secondary') button {
  /* applies to the "div[slot=secondary] button" element directly */ 
  /* ==> <button> x1 (Edit) */
}
</style>
```

arxpoetica:
@timhall (and @lukeed) there's a lot to unpack here, so let's just take it one step at a time.

How exactly is this a correct precompile translation:

```css
.component {
  :global {
    font: inherit;
    h1 { margin: 0 0 0.2rem; }
    h2 { margin: 0; }
  }
}
/* Transformed in preprocess using postcss-nested and postcss-global-nested to */
:global(.component) {
    font: inherit;
}
.component :global(h1) { margin: 0 0 0.2rem; }
.component :global(h2) { margin: 0; }
```

Wouldn't it compile to:

```css
.component :global { font: inherit; }
.component :global h1 { margin: 0 0 0.2rem; }
.component :global h2 { margin: 0; }
```

Or is their something special in `postcss-nested*` that compiles anything beginning with a `:` colon to parenthetical nestings?

timhall:
@arxpoetica Oops, fixed the transform in my comment, it should compile to what you listed above

hybridwebdev:
> Css vars via @lukeed:
> 
> ```
> <div class="demo">    
>     <button on:click={() => value++}>Increment</button>
>     <button on:click={() => value--}>Decrement</button>
> </div>
> 
> <script>
>     export let value = 0;
> </script>
> 
> <style>
>     .demo {    
>         padding: 10px;
>         border-radius: 4px;
>         -webkit-appearance: none;
>         font-style: italic;
>         /* this is dynamic */
>         color: {value % 2 ? green : red};
>     }
> </style>
> 
> 
> 
> <style>
>     .demo {    
>         padding: 10px;
>         border-radius: 4px;
>         -webkit-appearance: none;
>         font-style: italic;
>     }
> </style>
> 
> <div class="demo" style="color:green;">...</div>
> <div class="demo" style="color:red;">...</div>
> 
> 
> 
> <style>
>     .demo {    
>         padding: 10px;
>         border-radius: 4px;
>         -webkit-appearance: none;
>         font-style: italic;
>         color: var(--color);
> </style>
> 
> <div class="demo" style="--color:green;">...</div>
> <div class="demo" style="--color:red;">...</div>
> ```

JS variables inside style blocks is a TERRIBLE idea. The nightmare scenarios this would create in terms of debugging would just be astronomical. Better to keep the 2 separate. If you find a use case for something like this, then aphroditeJs or other CSS js frameworks would be more ideal.

Rich-Harris:
Closing this in favour of #13

PR #16, opened 2020-01-01
Prop Directives

lukeed:
[**Rendered**](https://github.com/sveltejs/rfcs/blob/prop-directives/text/0000-prop-directives.md)

### DIFF
--- a/text/0000-prop-directives.md
+++ b/text/0000-prop-directives.md
@@ -1,0 +1,118 @@
+- Start Date: 2020-01-01
+- RFC PR: [#16](https://github.com/sveltejs/rfcs/pull/16)
+- Svelte Issue: (leave this empty)
+
+# Props as Directives
+
+## Summary
+
+Should we allow a Component's directives to be dynamic/configured via an exported prop?
+
+## Motivation
+
+Specifically, the `transition:` directive sparked this RFC.<br>
+It would be amazing if we can allow Component consumers to configure transitions for the Components that allow them.
+
+However, I figure that once we allow `transition:` to be dynamic, it may implicitly require we allow all other directives (included or custom) to be dynamic, too.
+
+## Detailed design
+
+Let's assume I have an `<Alert/>` component that wants to run some kind of transition when it enters & leaves the page.
+
+For now, the `<Alert/>` will use a `fade` animation from the `svelte/transition` group, which (currently) is not available for configuration:
+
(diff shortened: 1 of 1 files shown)


pngwn:
Unless I'm gett confused, this is already mostly possible with a noop transition function ([REPL](https://svelte.dev/repl/f17c1b6791cc4eaf81457d6b25fa8970?version=3.16.7)).

The only addition we would need to make is to allow transitions to receive false or null instead of a function to disable transitions. The current noop approach incurs more cost than not trying to run the function at all, as far as I'm aware. The same is true for other directives that use functions, you can pass them as props ([REPL](https://svelte.dev/repl/aa3a412dcab4461881ca1c8a36586573?version=3.16.7)).

lukeed:
Closing this since as @pngwn pointed out, it's actually possible 🎉 

```html
<!-- Alert.svelte -->
<script>
	export let show = true;
	// NO transition by default
	export let transition = () => {};
</script>

{#if show}
<div transition:transition role="alert">
	<button type="button" class="close" on:click={onClose}>
		<span aria-hidden="true">×</span>
	</button>
	<slot/>
</div>
{/if}


<!-- Consumer.svelte -->
<script>
	import { Alert } from 'somewhere';
	import { fly } from 'svelte/transition';
	import { quintOut } from 'svelte/easing';

	let show = false;
	const transition = x => fly(x, {
		delay: 250,
		duration: 300,
		x: 100,
		y: 500,
		opacity: 0.5,
		easing: quintOut
	});
</script>

<Alert {show} {transition}>
	<p>Custom Alert that flies in/away on visibility toggle!</p>
</Alert>
```

---

At this point, I'd squabble about the syntax to make this a bit easier, but I don't think it's reason enough to keep this open or pursue at this time.

FWIW, I think something like this would be ideal:

```html
<!-- Alert.svelte -->
<script>
	export let show = true;
</script>

{#if show}
<!-- allow any transition, in/out, or animation directives -->
<div transition:* in:* out:* animation:*  role="alert">
	<button type="button" class="close" on:click={onClose}>
		<span aria-hidden="true">×</span>
	</button>
	<slot/>
</div>
{/if}

<!-- Consumer -->
<Alert transition:fly={{ .. settings .. }}>
	<p>Custom Alert that flies in/away on visibility toggle!</p>
</Alert>
```

PR #17, opened 2020-01-22
Pipes

skytracer:
[Rendered](https://github.com/skytracer/rfcs/blob/pipes/text/0000-pipes.md)

### DIFF
--- a/text/0000-pipes.md
+++ b/text/0000-pipes.md
@@ -1,0 +1,227 @@
+- Start Date: 2020-01-22
+- RFC PR: (leave this empty)
+- Svelte Issue: (leave this empty)
+
+# Pipes
+
+## Summary
+
+Pipes are a way to write display-value transformations that you can declare in your HTML.
+A pipe takes in data as input and transforms it to a desired output
+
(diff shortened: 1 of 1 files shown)


Conduitry:
Thanks for going through the process of an RFC for this, but there are a few reasons we wouldn't want to implement this feature in this way.

- It's a breaking change. Someone could conceivably be using bitwise OR in expressions already, and this would change the meaning of those.
- It's not javascript syntax. To avoid burdens with parsing and with other tooling, we need to keep the contents of the `<script>` tag and the expressions in the template to remain parseable as vanilla javascript, which `foo | bar : baz | quz` is not.
- It's a saving of only a couple keystrokes compared to the function call syntax
- It's less explicit and introduces a specialty way of doing something that already has a solution

It's currently a stage 1 draft so I wouldn't get too excited, but if something like the [pipeline operator](https://github.com/tc39/proposal-pipeline-operator) landed as an official feature in javascript itself, that'd probably be how something like this would be done - with no action on Svelte's part other than to update the version of the parser we're using.

PR #28, opened 2020-09-01
Sapper as middleware

ajbouh:
[**Rendered**](https://github.com/metered/rfcs/blob/master/text/0005-sapper-as-middleware.md)

### DIFF
--- a/text/0005-sapper-as-middleware.md
+++ b/text/0005-sapper-as-middleware.md
@@ -1,0 +1,165 @@
+- Start Date: 2020-08-31
+- RFC PR: 
+- Svelte Issue: 
+
+# Sapper as Middleware
+
+## Summary
+
+The Sapper build process should emit a mostly self-contained blob of javascript (and supporting files) that
+implements all server-side routes, SSR, and DOM code for the entire application. This javascript should
+export (at least) two functions:
+
+- A `middleware(req, res)` function that will serve a dynamic request.
+- A `assets(req, res)` function that will serve a static asset request.
+
+The supporting files *must* also include a manifest of files that can be uploaded to a CDN.
+
+This javascript *must not* start an Express server.
+
+## Motivation
+
+Working with Sapper in production for a large scale project has identified a number of
+challenges that arise from the unintended complexity of stale assumptions.
+
+By reducing these assumptions to the *absolute minumum*, we aim to:
+
+- Make it easier to deploy Sapper in efficient, scalable configurations
+- Make it easy to use Sapper applications in more scenarios than just a single standalone site
+- Make it easier to extend Sapper to suit specific project needs (without requiring modifications of Sapper itself)
+
+As of 0.28, the Sapper build process emits a completely standalone directory in `__sapper__/build`
+that starts an Express server listening on `$PORT` to serve the Sapper site.
+
+To do this Sapper makes many nuanced assumptions about the dependencies and runtime environment of
+applications it is building. Many of these assumptions are unnecessary sources of complexity during
+both developent and deployment.
+
+## Detailed design
+
+The general behavior of client-side bundling for Sapper should remain unchanged. Components are compiled for
+each route as SSR code and as DOM code. The behavior of the DOM (client-side) entry point is also unchanged.
+We expect the client-side bundle to be just as completely self-contained as it is today.
+
+This is in contrast with the server-side bundling operation. Instead of emitting a completely free-standing
+NodeJS script that starts an Express server, it would instead emit a partially bundled .js file that the user
+should include in a script of their own. This server-side .js file would look something like:
+
+```javascript
+
+export function middleware(req, res) {
+  // Sapper routing logic here to handle dynamic requests ...
+}
+
+export function assets(req, res) {
+  // sirv-based logic here to serve requests for static assets ...
+}
+
+```
+
+There must also be a manifest that's emitted to make deployment as straightforward as possible. A simple
+manifest might look something like:
+
+```json
+{
+  "assets": {
+    "directory": "assets",
+    "entries": {
+      "client/index.5bdaeb21.js": {
+        "headers": [
+          {
+            "key": "Cache-Control",
+            "value": "max-age=31536000, immutable"
+          },
+          {
+            "key": "Content-Type",
+            "value": "application/javascript"
+          }
+        ]
+      },
+      "index.css": {
+        "headers": [
+          {
+            "key": "Content-Type",
+            "value": "text/css"
+          }
+        ]
+      }
+    }
+  },
+  "functions": {
+    "index.js": {}
+  }
+}
+```
+
+Bringing everything together, the emitted directory structure should look something like:
+
+```
+assets/
+  client/
+    index.5bdaeb21.js
+  index.css
+index.js
+index.d.ts
+manifest.json
+```
+
+A few important details of this layout:
+
+- All static assets, including user-supplied (from `static/` directory) and Sapper-generated
+  (from compilation of Svelte components) are under a single directory prefix
+  (`assets/` in the example above)
+  
+- `manifest.json` is intended to only be used at deployment and development time, not *necessarily* at
+  run-time. (This is unlike `build.json` in Sapper 0.28, which is used only at run-time)
+  
+- Users must use/mount the exported `middleware(req, res)` function in an Express-compatible webserver.
+  The specific contract we expect for `req` and `res` parameters would be described by the types emitted
+  in `index.d.ts`.
+  
+  Compatibility libraries should be written to adapt popular serverless environments to these contracts.
+
+- Users are *not required* to use the exported `assets(req, res)` function in `index.js`. They can also
+  provide their own asset serving logic based on the information in `manifest.json`.
+
+## How we teach this
+
+Updating the existing template projects should probably be enough for simple projects. We should also
+add additional example projects that deploy Sapper applications in more scenarios.
+
+## Drawbacks
+
+Together, these changes represent a breaking change to Sapper's external contract. It's possible that
+Sapper *never* wants to support the developer experience that's outlined here.
+
+## Alternatives
+
+I've previously used a series of RegExps to pull routes out of the generated code a "manually" built this
+architecture with sed and shell script hackery.
+
+## Unresolved questions
+
+### Are people okay dropping Webpack support to make the above easier to implement?
+
+Support for Webpack increases the surface area for configuration and testing and may not be worth the trouble.
+
+### How satisfied are people with the current experience of `sapper dev`?
+
+Making Sapper emit middleware would definitely affect how reloading dev experiences operate!
+
+### What deployment targets do people care about?
+
+To start with, we expect Sapper apps to be trivially deployable to at least:
+
+- Cloudfront + AWS Lambda
+- Vercel
+- Vanilla NodeJS
+
+And hopefully also other targets like:
+
+- Cloud Functions for Firebase
+- Cloudflare Workers
+
+Others?
+


lukeed:
Looking good - I've always wanted Sapper to just be a middleware function.

Can you explain a bit about what `index.js` contains? And why there's `assets/*` vs `assets/client/*` directories?

Is `sirv` effectively re-exported as the `assets` function? 

Looking at the `manifest.json` – is there some new way to configure headers, especially `Cache-Control`? Where do these come from? (I get that built, versioned files can be assumed/given a Cache-Control, but what about others?)

ajbouh:
> Can you explain a bit about what index.js contains?

It contains something like this:

```typescript
export function middleware<Rq extends Req, Rs extends Res>(opts: MiddlewareOptions<Rq, Rs> = {}) {
	const { session, error_handler, context, ignore } = opts;

	return compose_handlers<Rq, Rs>(ignore, [
        	// ...

		dev && get_dev_handler(),

		get_server_route_handler({
			routes: manifest.server_routes, 
			error_handler
		}),

		get_page_handler(
			manifest,
			session || noop_session,
			context || noop_context,
		)
	].filter(Boolean) as Handler<Rq & Req, Rs & Res>[]);
}

export function assets<Rq extends Req, Rs extends Res>(opts: AssetsOptions<Rq, Rs> = {}) {
	return compose_handlers<Rq, Rs>(ignore, [
		// ...
		get_assets_handler<Rq, Rs>(),
	].filter(Boolean) as Handler<Rq & Req, Rs & Res>[]);
}
```

> And why there's assets/* vs assets/client/* directories?

`assets/` is basically a combination of the input `static` directory and the compiled `client/*` assets. Since these namespaces are layered during deployment I decided to keep things simple and demand that they're just copied to the same directory.

> Is `sirv` effectively re-exported as the assets function?

Yes, essentially. But with support for custom headers. More specifically it's something like:

```typescript
import sirv from 'sirv';
import {Stats} from 'fs';

import { dev, Req, Res, assets_manifest } from '@sapper/internal/manifest-server';

export function get_assets_handler<Rq extends Req, Rs extends Res>() {
	const entries = assets_manifest.entries
	return sirv<Rq, Rs>(assets_manifest.directory, {
		dev,
		setHeaders: (res: Rs, pathname: string, stats: Stats) => {
			for (const { key, value } of entries[pathname.substring(1)]?.headers || []) {
				res.setHeader(key, value)
			}
		}
	})
}
```

> Looking at the manifest.json – is there some new way to configure headers, especially Cache-Control? Where do these come from? (I get that built, versioned files can be assumed/given a Cache-Control, but what about others?)

At the moment they're essentially hard-coded during the build, but this is mostly because that's what Sapper was already doing. If there are other ideas about how to handle it that fit within the scope of this RFC I'm open to it.

Do you have any specific ideas?

lukeed:
> It contains something like this:

Gotcha, so it's just the generated output. Just making sure.

The rest sounds good and makes sense, thanks.

Maybe there can be an input manifest that can customize headers and whatever, but that'd be a separate RFC and not something to do now. Agree with your approach of just matching what's there -- just wanted to double-check that you weren't doing something more with those :)

text/0005-sapper-as-middleware.md
```diff
@@ -0,0 +1,165 @@
...
+          {
+            "key": "Content-Type",
+            "value": "text/css"
+          }
+        ]
+      }
+    }
+  },
+  "functions": {
+    "index.js": {}
```
benmccann, line 91: what would the right-hand side of this contain?

text/0005-sapper-as-middleware.md
```diff
@@ -0,0 +1,165 @@
...
+applications it is building. Many of these assumptions are unnecessary sources of complexity during
+both developent and deployment.
+
+## Detailed design
+
+The general behavior of client-side bundling for Sapper should remain unchanged. Components are compiled for
+each route as SSR code and as DOM code. The behavior of the DOM (client-side) entry point is also unchanged.
+We expect the client-side bundle to be just as completely self-contained as it is today.
+
+This is in contrast with the server-side bundling operation. Instead of emitting a completely free-standing
```
benmccann, line 44: > it would instead emit a partially bundled .js file that the user should include in a script of their own

For the standard Express/Polka case that we support today, would the user do a second bundling operation or their own script could be run directly?

text/0005-sapper-as-middleware.md
```diff
@@ -0,0 +1,165 @@
...
+            "key": "Cache-Control",
+            "value": "max-age=31536000, immutable"
+          },
+          {
+            "key": "Content-Type",
+            "value": "application/javascript"
+          }
+        ]
+      },
+      "index.css": {
```
benmccann, line 80: I'm a bit confused by this example because it doesn't quite correspond to what I'd normally see in a Sapper application. Is this the css corresponding to the JS file above? If so, I'd expect it to be named something like `client/assets/index.34fa2382.css`. Or is this something more like `global.css`?

text/0005-sapper-as-middleware.md
ajbouh: Probably details about the requirements of this function.

Things like environment variables, minimum memory, required node versions.

text/0005-sapper-as-middleware.md
ajbouh: It would be runnable directly without additional node dependencies, but if they were previously depending on sapper's bundling to make their server code easier to distribute they would need to switch to doing their own explicit bundle operation.

text/0005-sapper-as-middleware.md
ajbouh: Yeah this is like global.css. I changed global.css to index.css in the base project template I use and I keep forgetting that everyone else calls it global.css.

benmccann:
I think it's important to understand what this looks like end-to-end on a couple of the major platforms. It's hard to evaluate whether this file format is necessary and/or helpful without really seeing how it will be used. E.g. the biggest problem for serverless deployments is how do you split the routes to build separate bundles and that's not really discussed here. Nor is the tooling that would consume this file discussed. It seems like it'd be really helpful to figure out how to build them and make that work on a couple platforms and then look at what the hard parts are and what can and should be shared.

Headers for static assets are already calculated today. To me the main question here is where that should be done. In Sapper or in the deployment tooling?

Advantages of doing it in Sapper:
* If Sapper changes its output directory you probably don't need to update your deployment tooling
* Allows you to specify configuration in a single place if deploying to multiple platforms and you want the same config in all of them

Advantages of doing it in deployment tooling:
* Can support configuration values or formats that are only present on a single platform
* Existing deployment tooling can be utilized. E.g. if AWS, Vercel, etc. have deployment tooling it can be used without making it understand this new file format
* Easier to override headers for specific files. The deployment tooling typically already has this built in. We would need some way of specifying this config value to Sapper and then passing it into the file. This extra surface area to maintain that I'm not sure we need to given that the deployment tooling already has it
* Easier to have different configuration per platform if deploying to multiple platforms. You only need to know how to configure the platform you're deploying to instead of the platform and Sapper

It's not clear to me that this file really helps all that much. E.g. if I look at the implementation of `vercel-sapper` setting up the static assets is only a few lines:

https://github.com/thgh/vercel-sapper/blob/b7d0a050d5e23d64c7e72e11847acc60311d4921/index.js#L98

I'm fully supportive of making it easier to deploy to multiple platforms. But overall, I don't understand the need for list all the static assets in a new file. This seems like a non-problem that's being solved in just a few lines of code today

ajbouh:
> E.g. the biggest problem for serverless deployments is how do you split the routes to build separate bundles and that's not really discussed here. Nor is the tooling that would consume this file discussed. It seems like it'd be really helpful to figure out how to build them and make that work on a couple platforms and then look at what the hard parts are and what can and should be shared.

I respectfully disagree with this claim. I've done a number of serverless deployments to both Vercel (back when it was called Now) and to AWS Lambda@Edge. This RFC is a direct result of my experiences and efforts to deploy Sapper in real world serverless environments. The biggest challenges have been:

- How to identify which routes are dynamically served and which routes are static assets that should be offloaded to a CDN.

- How to keep the packaged serverless function as small as possible. Part of this means *not* including static assets in the packaged serverless function.

- How to match the serverless platform's internal request lifecycle logic to Sapper's. For example, AWS Lambda provides the request body as a base64-encoded blob.

- How to avoid needing an entire Express server listening on a localhost port (and the associated overhead of using it for serialization and deserialization). I was very surprised to learn that this is something that Vercel *secretly adds* to your project when you deploy there.

- How to handle server-side configuration without access to process.env (which is not available in AWS Lambda@Edge)

> ...

> Headers for static assets are already calculated today. To me the main question here is where that should be done. In Sapper or in the deployment tooling?

The idea of generating headers *within* Sapper is part of the current implementation. I personally view it as out-of-scope for this particular RFC to change this. Is there a lens should we consider this RFC through that would imply that it's actually within scope?

> ...

> - Can support configuration values or formats that are only present on a single platform

You can of course add your own configuration in platform-specific tooling. This strikes me as a sort of false dichotomy. The real question is how much of Sapper's internal implementation details should these deployment tools be dependent on?

> - Existing deployment tooling can be utilized. E.g. if AWS, Vercel, etc. have deployment tooling it can be used without making it understand this new file format

In my experience this isn't true. For example, to deploy to AWS I've had to write RegExps that scrape routes from internal code generated by Sapper. It's much harder to reverse engineer Sapper's code generation logic and write these RegExps than it is to parse a JSON file that presents a well thought out public interface.

The existing effort in vercel-sapper (and the lack of deployment tooling for other platforms) should be an indication that the existing approach of doing things in the deployment tool requires uncommon expertise in *both* Sapper's output formats *and* the particulars of that deployment platform. The current approach is also a source of community inertia and a constraint on the evolution of Sapper itself.

> - Easier to override headers for specific files. The deployment tooling typically already has this built in. We would need some way of specifying this config value to Sapper and then passing it into the file. This extra surface area to maintain that I'm not sure we need to given that the deployment tooling already has it

Having a manifest file does not preclude overriding or adding headers in your specific deployment tool though, right?

> - Easier to have different configuration per platform if deploying to multiple platforms. You only need to know how to configure the platform you're deploying to instead of the platform and Sapper

This isn't strictly true. It's important to understand *both* the implementation details of Sapper *and* the deployment platform and decide how to properly integrate them. IMO the core issue today is that Sapper's directory layout, its headers and routing policy are a large, brittle, and under-specified accidentally public interface that is fragile to any number of internal Sapper changes.

> ...

> It's not clear to me that this file really helps all that much. E.g. if I look at the implementation of vercel-sapper setting up the static assets is only a few lines:
>
> https://github.com/thgh/vercel-sapper/blob/b7d0a050d5e23d64c7e72e11847acc60311d4921/index.js#L98
>
> I'm fully supportive of making it easier to deploy to multiple platforms. But overall, I don't understand the need for list all the static assets in a new file. This seems like a non-problem that's being solved in just a few lines of code today

It's true that the headers alone are specified in just a few lines, but it's deceptive to think that support for Sapper's routing policy is sufficiently specified by just those few lines. Most of that entire ~120 line file is spent on trying to massage the Sapper output directory structure into something that works for Vercel's approach to hosting and routing.

A simple question to evaluate this claim: if Sapper were to completely rearrange its output (or add headers) how much, if any, would the downstream tools like vercel-sapper need to change?

Users of vanilla NodeJS would only need to change `node __sapper__/build` to something like `node __sapper__/output`. That's pretty good.

Users of vercel-sapper would not be so lucky. The PR made to vercel-sapper would probably be much larger and more intricate that just editing the few lines you've linked to.

If, however, vercel-sapper were to be interpreting a manifest instead of attempting to interpret the implementation details of Sapper's raw directory structure, I expect no change would be needed at all. (Assuming of course that the schema and location of the manifest were not changed.)

benmccann:
I agree with all of the issues you listed. However, the RFC doesn't really describe how the majority of them will be addressed by this proposal, which is my overarching concern. I'm open to the idea that this may help, but it's not really connecting the dots for me

> Users of vercel-sapper would not be so lucky. The PR made to vercel-sapper would probably be much larger and more intricate that just editing the few lines you've linked to.

Can you explain? There are two lines that refer to `__sapper__/build` that would need to be updated to `__sapper__/output`. I agree with you that your proposal would help here by avoiding a need for a change and included that in my list of pros and cons. It doesn't weight super heavily for me because this directory structure isn't something that we've ever changed before to my knowledge and it doesn't seem like a hard fix if it does need to change. My concern is that this is the only problem that's really obvious to me that's being solved, but it seems like such a small one. You listed a lot of other issues that seem much bigger and I think any solution should describe those issues in more detail and how they are solved.

(As a side note, I think if we wanted to specify the static assets we could do it far more concisely by specifying the directories or a glob rather than the list of their contents, and that would probably make me happier)

text/0005-sapper-as-middleware.md
benmccann: Is that something I want Sapper to control though? How would it know what to put there? It seems like something I need to specify as a developer

pngwn:
On this point. 
> How to handle server-side configuration without access to process.env (which is not available in AWS Lambda@Edge)

Luke’s recent PR could go some way to addressing that in user code at least given that is relies on a build mode and isn’t tied to any particular environment.

ajbouh:
Link?

pngwn:
I meant RFC sorry. Thinko.

ajbouh:
I think Luke's RFC works best in circumstances where there's only one value for a given configuration setting and that value is known at the time of svelte component compilation time. Right?

When deploying to multiple regions (like, IMO, serverless is uniquely suited to do) you need some bits of configuration to be specific to the region that you're deploying in. Without doing this you end up having a function in one region (say, Asia) that will end up depending on services in another, different region. This would negatively affect both performance and reliability.

So far for AWS Lambda I've been using AWS SSM Parameter Store for my server-side configuration values and a sed/gsub/replace-style approach during packaging to hardcode the SSM Parameter Store path prefix to the AWS region that I'm deploying the function to. Not an approach I love, but is required to avoid accidental cross-region dependencies.

benmccann:
> The goal of this RFC is not to address everything needed make Sapper deployable in serverless contexts

If this doesn't address serverless then I'm not sure what the goal is. The motivations section says it makes it easier to deploy and extend Sapper, but doesn't give examples of the specific scenarios it's intended to help with. I'm not really sure how we can talk about deployment generally and ignore the elephant in the room of serverless since that's what the vast majority of user issues around deployment are for.

This might be our fault for not specifying the RFC process well enough. I get that what I'm asking for is more work, but we need to think about solutions holistically. It's very hard to evaluate a proposal if I'm seeing only a small part of the solution. I'm not saying everything has to be 100% fleshed out and I don't want to make the bar so high that RFcs are unapproachable. At the same time, if I can't understand how this file is generated or read then I can't understand if it's the best solution in the larger context.

> If Sapper did something more sophisticated with routing (for example route splitting) then the changes to vercel-sapper would have to be much more involved. 

Let's focus on concrete issues being solved instead of "ifs" and hypotheticals. Is this a problem this RFC is meant to address or not? What's the primary motivator or a list of representative problems we can talk concretely about?

>  It also explicitly enumerates the assets in that directory, to avoid situations where there might be other files in that directory that should not be uploaded or served.

That doesn't exist today. If there's a scenario where it's required let's explicitly focus on that as a motivation for this change. As of right now, I can't understand what problem is being solved

ajbouh:
> > The goal of this RFC is not to address everything needed make Sapper deployable in serverless contexts
> 
> If this doesn't address serverless then I'm not sure what the goal is. The motivations section says it makes it easier to deploy and extend Sapper, but doesn't give examples of the specific scenarios it's intended to help with. I'm not really sure how we can talk about deployment generally and ignore the elephant in the room of serverless since that's what the vast majority of user issues around deployment are for.

The goal of this RFC is very narrow. It's to make Sapper work like normal middleware. That's all.

Given the state of the Sapper ecosystem, I don't think I'm in a position to completely design and define an entire solution for Sapper serverless deployment and then lead the discussion. I have an implementation that works for my needs but am beginning to suspect that I am too far ahead of the curve. Perhaps we should wait for others to hack their way through the current implementation details of Sapper and study these approaches once there are enough of them.

> This might be our fault for not specifying the RFC process well enough. I get that what I'm asking for is more work, but we need to think about solutions holistically. It's very hard to evaluate a proposal if I'm seeing only a small part of the solution. I'm not saying everything has to be 100% fleshed out and I don't want to make the bar so high that RFCs are unapproachable. At the same time, if I can't understand how this file is generated or read then I can't understand if it's the best solution in the larger context.

I understand the desire to think about this holistically, but I don't think the solution is to have such a large scope for a single initial RFC. In my opinion, before we can have a useful discussion about a holistic solution to serverless deployment, we need more members of the community involved in the requirements gathering effort.

> > If Sapper did something more sophisticated with routing (for example route splitting) then the changes to vercel-sapper would have to be much more involved.
> 
> Let's focus on concrete issues being solved instead of "ifs" and hypotheticals. Is this a problem this RFC is meant to address or not? What's the primary motivator or a list of representative problems we can talk concretely about?

The point of the manifest, very simply, is to eliminate the coupling between external tools that *use* Sapper's output from Sapper's own *internal* layout decisions. `vercel-sapper` is a big offender on this front. If other tools emulate the approach that `vercel-sapper` takes, it will become harder and harder for Sapper to make changes that don't break other things in the ecosystem.

> > It also explicitly enumerates the assets in that directory, to avoid situations where there might be other files in that directory that should not be uploaded or served.

> That doesn't exist today. If there's a scenario where it's required let's explicitly focus on that as a motivation for this change. As of right now, I can't understand what problem is being solved

Except this does exist today. This happens all the time on macOS with .DS_Store files. The current "workaround" to this sort of thing is the use of ignore files, which are a needless source of confusion and complexity. See https://twitter.com/Rich_Harris/status/1270871878018699264 for a timely example. It would never be an issue if there was an explicit file manifest.

benmccann:
> In my opinion, before we can have a useful discussion about a holistic solution to serverless deployment, we need more members of the community involved in the requirements gathering effort.

I think we've had probably had enough requests come in that we have a good idea what the biggest requirements are. Check out the middleware bucket here: https://github.com/sveltejs/sapper/projects/1

> The goal of this RFC is very narrow. It's to make Sapper work like normal middleware. That's all.

I'm not sure why a new file format is required to allow that, which is why it looked to me like this was trying to do more.

Today the user creates a `server.js`:
```
polka() // You can also use Express
	.use(
		sirv('static', { dev }),
		sapper.middleware()
	)
	.listen(PORT, err => {
		if (err) console.log('error', err);
	});
```

So we already have a sapper middleware that can be used. I'm not really sure what it means to allow using Sapper as a middleware that you can't already do. What's stopping you from using something besides Polka in this file?

ajbouh:
> > In my opinion, before we can have a useful discussion about a holistic solution to serverless deployment, we need more members of the community involved in the requirements gathering effort.
> 
> I think we've had probably had enough requests come in that we have a good idea what the biggest requirements are. Check out the middleware bucket here: https://github.com/sveltejs/sapper/projects/1

Thanks for the pointer. I don't see much here about the various other serveless deployments beyond Lambda. I'd expect to see comments or interest about:

- Vercel
- Cloudflare Workers
- Azure Functions
- Firebase Functions

I'd also expect to see things raw qualitative or quantitative information about:
- experience with cold starts
- server-side application bundle sizes
- the possible value of route splitting

> 
> > The goal of this RFC is very narrow. It's to make Sapper work like normal middleware. That's all.
> 
> I'm not sure why a new file format is required to allow that, which is why it looked to me like this was trying to do more.
> 
> Today the user creates a `server.js`:
> 
> ```
> polka() // You can also use Express
> 	.use(
> 		sirv('static', { dev }),
> 		sapper.middleware()
> 	)
> 	.listen(PORT, err => {
> 		if (err) console.log('error', err);
> 	});
> ```
> 
> So we already have a sapper middleware that can be used. I'm not really sure what it means to allow using Sapper as a middleware that you can't already do. What's stopping you from using something besides Polka in this file?

The current .js files emitted by `sapper build` all make assumptions about where client assets are. It assumes that generated **service worker and client assets** will be on disk and served by the function returned by `sapper.middleware()`. This is not an appropriate use of compute or bandwidth resources. Much better is to serve static files directly from your CDN.

With the current approach, how do you know which files to upload to S3? How do you know which headers to give those files? How do you know which routes (or route prefixes) to serve those files on? If there's a conflict between a static asset route and a dynamic server route, which should win?

Not only do I think we should avoid *writing code* that assumes Sapper will be running inside an Express app on localhost, I  think we should avoid assuming it mentally as well.

Along these lines, Sapper today also:
- assumes it is responsible for bundling your entire server-side application. This isn't particularly helpful and creates unnecessary challenges when operating outside a vanilla NodeJS and Express-like world
- assumes you are running an Express server available at `http://127.0.0.1:${process.env.PORT}`.

benmccann:
So is the primary thing we're trying to enable here serving static files from a CDN or S3 then?

I know Antony has used Vercel's Smart CDN and others have used Cloudflare. I've never used S3 as a CDN. If there are specific issues with that one, it would be helpful to include details about it in the RFC.

If this is the problem you're trying to solve, then details I would like to see:
* What is an example of a CDN you can or can't serve from today? What are the difficulties with serving from the CDN you're attempting to use?
* Do you need to remove the Sapper middleware's serving of `client` files if you want the CDN to do it?
* Does the asset path need to be configurable to support a CDN? Can you link to the relevant section of docs from the CDN you're using so that we can familiarize ourselves with the issue that you're facing?
* Does the CDN you're trying to use require the paths to images and other assets to be changed in some way? Can you explain how? E.g. do we need to change the base path of images, but not links? How would we do that?
* What type of documentation should be added to support users that want to serve from a CDN?

ajbouh:
> - What is an example of a CDN you can or can't serve from today? What are the difficulties with serving from the CDN you're attempting to use?

I feel like I'm a broken record on this topic...

Questions you need to answer to run on AWS S3 + AWS Cloudfront + AWS Lambda:

- Which files do you upload as static assets?
- Which headers should you give those files?
- Which routes (or route prefixes) to serve those files on?
- Which routes are dynamically served and which routes are static assets that should be offloaded to a CDN?
- If there's a conflict between a static asset route and a dynamic server route, which should win?
- What is the minimum set of files that are required in order to serve a dynamic request? (This means not including static assets in the packaged serverless function.)

> - Do you need to remove the Sapper middleware's serving of client files if you want the CDN to do it?

Yes, most serverless platforms have hard lambda size limits. Even below this limit, the larger your lambda is the slower and more expensive your hosting is.

> - Does the asset path need to be configurable to support a CDN? Can you link to the relevant section of docs from the CDN you're using so that we can familiarize ourselves with the issue that you're facing?
> - Does the CDN you're trying to use require the paths to images and other assets to be changed in some way? Can you explain how? E.g. do we need to change the base path of images, but not links? How would we do that?

No you don't need any of this. You just need to know what Sapper is expecting the route mapping to be.

> - What type of documentation should be added to support users that want to serve from a CDN?

I would expect people to use an ecosystem of separate tools to deploy their Sapper application. These tools would have their own documentation, in the vein of `vercel-sapper`. The manifest I've described is the interface to those tools. The documentation that Sapper would be responsible for would be to document the parts of its output that can be reasonably expected to remain backward compatible. I expect the manifest would be the most important part of this documentation.

benmccann:
>> Do you need to remove the Sapper middleware's serving of client files if you want the CDN to do it?

> Yes

Ok, that's probably the most user-impacting change here, but one I support. I think it's inconsistent and confusing that Sapper has two different static directories and one is user-configured and one is included in the middleware.

What does that look like for Polka/Express users? I sort of think it'd make sense to just remove that from Sapper and then change the template to look more like this:

```
import sirv from 'sirv';
import path from 'path';
import polka from 'polka';
import * as sapper from '@sapper/server';

const { PORT, NODE_ENV } = process.env;
const dev = NODE_ENV === 'development';

const clientPath = path.join('__sapper__', dev ? 'dev' : 'build', 'client');
const clientAssets = sirv(clientPath, dev ? {} : {
  	maxAge: 31536000, // 1Y
  	immutable: true
});

polka() // You can also use Express
	.use('client', clientAssets)
	.use(
		sirv('static', { dev }),
		sapper.middleware()
	)
	.listen(PORT, err => {
		if (err) console.log('error', err);
	});
```

Or probably better still would be to take https://github.com/sveltejs/sapper-template/pull/248 to its conclusion by hashing all assets and removing the need to serve the `static` directory at all. It makes for an easier deployment and configuration story to have only a single static directory instead of two and then the template is no more complicated than it is today.

ajbouh:
But you can't remove the `static/` directory from Sapper because it needs to know what routes must skip the service worker.

These suggested header values for the client directory would also be very bad for service-worker.js file(s), which require special cache-control headers in order to avoid being wedged forever at some old version.

ajbouh:
Hmm, unless you're actually suggesting we drop the service worker?

If we're looking to keep the service worker then there's a lot of nuance needed to properly balance static and dynamic routes. This nuance is exactly the reason I think that Sapper should be explicit about both headers and assets.

benmccann:
Oh, yeah, I always forget about the service worker because I don't use it. I actually would be in favor of removing it from Sapper. But assuming we keep supporting it, I would think we should move it out of the middleware as well. I updated the example above to demonstrate that. It's getting a bit verbose now, but I think there are some utility functions we could provide to simplify it.

ajbouh:
I don't think I understand your counter proposal. [According to Rich](https://github.com/sveltejs/sapper/issues/660):
> Given that the static folder has special meaning (e.g. it's used to determine what should go in the service worker manifest) it probably makes sense just to include sirv inside Sapper, rather than having it in the project template separately.

Broadly speaking, it sounds to me like Sapper doesn't yet have a principled policy on what constitutes a stable public interface for other tools to interact with.

Without a common understanding of what the existing public stable interface is intended to be, it's probably too difficult for us to talk about anything that would impact it.

My default here has been to preserve Sapper's freedom to change the answers to the questions I posed above (what files do we upload, what routes are they served on, etc). Specifying a manifest is the simplest way I have thought of to do this, but I'm open to alternatives.

But maybe folks don't think this freedom is worth the cost of exposing a "public" manifest?

If that's the case, then I would suggest the Sapper team provides an update to the official documentation that includes:
- where different files will be written by `sapper build`
- a hosting contract for the various files (routes, headers, dynamic vs static precedence, etc.)
- some sort of a commitment not to change the file layout or the hosting contract in the near future

I can then update this RFC with a concrete set of implied changes to that public documentation, if any

benmccann:
I'm coming at this from another angle, which is that I think the user should have control over most decisions. Sapper's a bit of a monolith, but we've been talking a lot about splitting out various components like routing, css handling, etc. and making Sapper more of a collection of libraries and plugins. In that world, I could imagine that that we'd have a standalone Rollup plugin for generating the service worker manifest.

I think it can be a bit restrictive that the service worker manifest today contains exactly the files that are in the static directory. I could imagine the user might want to configure this a number of different ways. What we have today is as good a default as any, but we should make it easy to customize for your specific use case, which I think is really hard today. Here's a number of requests I've seen for customizing it:

* https://github.com/sveltejs/sapper/issues/983
* https://github.com/sveltejs/sapper/issues/667
* https://github.com/sveltejs/sapper/issues/685
* https://github.com/sveltejs/sapper/issues/625
* https://github.com/sveltejs/sapper/issues/527
* https://github.com/sveltejs/sapper-template/issues/60

> where different files will be written by sapper build

We basically have this: https://sapper.svelte.dev/docs#Deploying_from_sapper_build

> a hosting contract for the various files (routes, headers, dynamic vs static precedence, etc.)

I don't think Sapper should control this, but rather the user should. There's really nothing Sapper-specific about any of these values. There are perhaps some best-practices that are worth sharing. E.g. I think it's a good idea to hash static assets and give them long-lived cache expiration. But if you want to do something else and have a good reason for it I don't see why we should stand in your way.

> some sort of a commitment not to change the file layout or the hosting contract in the near future

We have a policy of documenting any breaking changes in the migration guide. I don't expect this to change, but if it did we would only do so with a major release, document it, and make sure people were aware of it.

ajbouh:
These are all strong points to make RE: service workers, routing, css, etc. but they're all far beyond the intended scope of this RFC.

Right now I'm just trying to preserve the existing Sapper contract. If in the future we want to simplify the contract and say "do anything you want for serving the files in this directory!" that's fine. That'd even be backwards compatible with this RFC since it's technically a relaxation of the contract that this RFC proposes.

Users are of course free to do anything they want no matter what the documentation says. 😄 

> We basically have this: https://sapper.svelte.dev/docs#Deploying_from_sapper_build

> You will need the `__sapper__/build` and static directories as well as the production dependencies in node_modules to run the application. Node production dependencies can be generated with npm ci --prod, you can then start your app with:

Right, which doesn't say anything about where the asset files are located within `__sapper__/build`, what routes to serve them on, what headers to serve them with or anything that resembles the suggested list of questions I'd need answered to understand how to frame the RFC in the context of Sapper's existing intentions for its public stable interface.

If there is enough dissatisfaction with the way Sapper currently handles asset headers and service worker manifests, I can withdraw this RFC until that's settled and we can revisit it once people are happy with how Sapper currently operates.

ajbouh:
> > a hosting contract for the various files (routes, headers, dynamic vs static precedence, etc.)
> 
> ... There's really nothing Sapper-specific about any of these values.

This just isn't true. Sapper has lots of hard coded ideas about how the files prefixed by `/client` are named, arranged, and cached. It also makes assumptions about both the service worker's location and about the precedence of static versus dynamic routes. Because of the way caching works, Sapper locks you into a number of choices for these that you aren't free to easily change between releases of your application.

All of these are open to design evolution of course, but those changes should be properly motivated and given their own discussion forum.

> > some sort of a commitment not to change the file layout or the hosting contract in the near future
> 
> We have a policy of documenting any breaking changes in the migration guide. I don't expect this to change, but if it did we would only do so with a major release, document it, and make sure people were aware of it.

Right, but we need to be specific about what constitutes a "breaking change". If `vercel-sapper` stops working, is that a breaking change or a bug in `vercel-sapper`? What about some other random project on npm that makes assumptions about the internal layout of Sapper's generated code? How do we decide which projects like `vercel-sapper` get to block Sapper releases?

benmccann:
I don't know that we've put so much thought into things that we want to ingrain everything as is. I'm all for documenting public APIs, but when we do so we should first make sure it's the API we want to have. Very often we've avoided documenting things that we're not in love or just haven't had much time to think about yet with because we're not sure it's not the contract we want to make. I think one of the major reasons we're not 1.0 yet is because there are things we know we want to change.

text/0005-sapper-as-middleware.md
```diff
@@ -0,0 +1,165 @@
...
+Working with Sapper in production for a large scale project has identified a number of
+challenges that arise from the unintended complexity of stale assumptions.
+
+By reducing these assumptions to the *absolute minumum*, we aim to:
+
+- Make it easier to deploy Sapper in efficient, scalable configurations
+- Make it easy to use Sapper applications in more scenarios than just a single standalone site
+- Make it easier to extend Sapper to suit specific project needs (without requiring modifications of Sapper itself)
+
+As of 0.28, the Sapper build process emits a completely standalone directory in `__sapper__/build`
```
thgh, line 31: I may have missed what changed in 0.28, do you have a link to the relevant PR that changed the build output?

thgh:
I have a question: if Sapper outputs a JS file that exports middleware, which program will build the express server? Or does this RFC just adds one extra file to the build output?

benmccann:
Thanks @jonatansberg !

> The biggest problem is the tight coupling towards Express/Polka and the filesystem (and more specifically the Node http and fs modules.)

I'm super interested that you looked into this a bit and have some ideas on improving it. I just saw your message on Discord, so will chime in there, but very exciting to have someone looking at it!

> No one is suggesting that Sapper should be concerned with the specifics of how you provision the required infrastructure on AWS.

I think it was suggested that Sapper include that in it's generated file, which is probably my biggest concern: https://github.com/sveltejs/rfcs/pull/28#discussion_r481780710

I'd really like to see more details about what would be put there, how it'd end up there, and how it would be consumed. One example of how this would work end-to-end would help a lot. I don't intend that to be a big burden, so it doesn't have to be a fully fleshed out system or that everything discussed here be implemented. Just enough to help understand and discuss this piece of the puzzle in the larger context

> Finally, there is also some relevant prior art in FAB (GitHub project). 

Cool! Thanks for sharing. I like the idea of aligning on some existing standard a lot. It seems like a young project that's not super widely adopted yet, but still might be better than going off on our own. [It looks to me like it supports S3, Cloudflare, and Lambda](https://fab.dev/guides/deploying) at the moment. Ideally this would be pluggable so that if there's a competing standard someone wants to use they could turn off the FAB Rollup plugin or whatever and enable their own. But I think it'd be nice to provide a good default and FAB's probably the best I've seen so far

text/0005-sapper-as-middleware.md
jonatansberg: Maybe just the search path would be enough to start with? I agree with @benmccann that platform specific requirements like CPU, memory, etc should come from elsewhere.

text/0005-sapper-as-middleware.md
ajbouh: No, this is bad wording. I'm trying to say that it's true in the current version and anchor it to a specific point in time so that someone looking back can know what "current" referring to.

ajbouh:
> I have a question: if Sapper outputs a JS file that exports middleware, which program will build the express server? Or does this RFC just adds one extra file to the build output?

So technically today user code is always the entry point, since users have their own server.js. The proposal here is to stop bundling that entry point and instead just generate a (mostly) standalone library that their code imports.

Users would be responsible for mounting and starting the express/polka server on their own. They'd also be responsible for doing any server side bundling they wanted on their own.

ajbouh:
Just saw these edits to your comment above, @benmccann:

> > But maybe folks don't think this freedom is worth the cost of exposing a "public" manifest?

> A manifest sort of makes sense, but I'm not sure Sapper should generate it because it takes away user control. Maybe it generates it only once or only adds missing entries, but a lot of it seems like it needs to be hand maintained. I just looked up an example for aws that's below. How would Sapper know how much memory you want, what the timeout should be, what concurrency you want to provision, etc.?

The manifest is solely intended to answer questions outlined in [this comment](https://github.com/sveltejs/rfcs/pull/28#issuecomment-686103506).

This is the minimum set of information you need in order to not depend on Express/Polka/Sirv.

I don't think this manifest should contain information that Sapper does not have an opinion of. Please only consider the details that are present in the example. I am not advocating for Sapper to be an expert in AWS concepts or to dictate values in any specific deployment provider's manifest.

I think we should keep the conversation here focused on the specifics of this RFC. This will help ensure this discussion is as approachable as possible.

benmccann:
I'm not trying to go off-topic, but it's a little hard to understand what exactly is in this proposal. I'm confused about this part specifically:

```
"functions": {
    "index.js": ?
}
```

I don't think you can say "Please only consider the details that are present in the example." and then not provide a full example. You had said "Things like environment variables, minimum memory, required node versions." would go there and that's what I was basing my comment off of. It's just too hard to have a conversation about it right now without more details. I think everyone is imagining this working a different way since there's not a lot of specificity, which is why we're going in circles.

ajbouh:
In this RFC it's just an empty object, exactly as given in the example.

In the future there will other things to go there (perhaps things from the list I outlined), but for now it's reserved for backwards compatibility purposes. That's why in the example code it's listed as just `{}`.

benmccann:
Ah, it was unclear to me that was provided only for forwards compatibility. Another way to accomplish that might be to do something like this:

```
"functions": [
  { path: "index.js" }
]
```

I think that has the advantage of being a lot clearer in the short-term when there are no other properties you want to specify. The two downsides would be that it's slightly more verbose and you can't do an indexed lookup on the path anymore (though I'm not sure that's a requirement anyway)

> The proposal here is to stop bundling that entry point and instead just generate a (mostly) standalone library that their code imports.
>Users would be responsible for mounting and starting the express/polka server on their own. They'd also be responsible for doing any server side bundling they wanted on their own.

Can you add some details about this to the RFC? I'm guessing we'd need to make some change to https://github.com/sveltejs/sapper-template at a minimum

jonatansberg:
If we want to make this backwards compatible we could do so by refactoring Sapper to remove the Node specific dependencies from the rendering path. Just introducing the lower level APIs I mentioned previously would go a long way. We could then wrap them to provide the same higher level express/connect middleware API we have today, meaning that existing projects would not break.

ajbouh:
> Ah, it was unclear to me that was provided only for forwards compatibility. Another way to accomplish that might be to do something like this:
> 
> ```
> "functions": [
>   { path: "index.js" }
> ]
> ```
> 
> I think that has the advantage of being a lot clearer in the short-term when there are no other properties you want to specify. The two downsides would be that it's slightly more verbose and you can't do an indexed lookup on the path anymore (though I'm not sure that's a requirement anyway)

I don't think adding the requirement that the functions be iterated over to accumulate a "path -> config" map is worth the effort but I could be wrong.

> 
> > The proposal here is to stop bundling that entry point and instead just generate a (mostly) standalone library that their code imports.
> > Users would be responsible for mounting and starting the express/polka server on their own. They'd also be responsible for doing any server side bundling they wanted on their own.
> 
> Can you add some details about this to the RFC? I'm guessing we'd need to make some change to https://github.com/sveltejs/sapper-template at a minimum

Yes, changes to the sapper-template would make sense, although the value of using a template goes down in this world, since you don't need nearly as complicated of a project configuration to use sapper as middleware.

ajbouh:
> If we want to make this backwards compatible we could do so by refactoring Sapper to remove the Node specific dependencies from the rendering path. Just introducing the lower level APIs I mentioned previously would go a long way. We could then wrap them to provide the same higher level express/connect middleware API we have today, meaning that existing projects would not break.

Yes, if people wanted this RFC to happen then I think we could figure out some type of upgrade path for most existing projects. Although this RFC advocates for no longer using rollup to bundle server dependencies. So over time that use case might end up rotting or generally becoming a source of user support requests.

benmccann:
I'm going to go ahead and close this since SvelteKit has adapters for different platforms which sounds like what we were trying to solve for here. We're not going to be adding any major new features to Sapper at this point. If there's anything missing from SvelteKit's adapters feel free to raise a new issue or RFC that talks about what gap exists in SvelteKit and we can take a look at closing that gap

PR #30, opened 2020-09-03
Add more details to encourage successful RFCs

benmccann:
We've gotten a good handful of RFCs over the past few days (woohoo!) and several people have commented both as authors and reviewers that they think we could improve things a bit. I think two main things have come up:
* people feel like they have to invest a lot into designing a solution, but really a lot of the value is in defining the problem well, gathering input from multiple parties, etc. The solution itself isn't always complicated, but aligning around it can be
* there's not a lot of focus on the content around the design. If someone just drops a design, it can be hard to tell what problems are being solved, how they arrived at this solution vs others, etc.

I tried to be a guinea pig for the new section here: https://github.com/benmccann/rfcs/blob/css/text/0000-sapper-css.md

### DIFF
--- a/0000-template.md
+++ b/0000-template.md
@@ -51,15 +51,15 @@ understand, and for somebody familiar with the implementation to implement. Any
 > new terminology should be defined here.
 
 > Explain not just the final design, but also how you arrived at it. What
-> constraints did you face? How does this deal with corner cases?
+> constraints did you face? Are there corner cases you've come up with solutions for?
 
-> Explain how your design fits into the larger picture. What are the other open
-> problems in this area? How does this design fit with potential solutions for those
-> issues?
+> Explain how your design fits into the larger picture. Are the other open problems
+> in this area you're familiar with? How does this design fit with potential
+> solutions for those issues?
 
 > Connect your design to the motivations you listed above. When describing a part of
-> the design, share an example of what it would look like to utilize the solution to
-> solve the problem.
+> the design, it can be useful to share an example of what it would look like to
+> utilize the implementation as solution to the problem.
 
 ## How we teach this
 
--- a/0000-template.md
+++ b/0000-template.md
@@ -17,11 +17,13 @@
 > it easier to do X" then give an example of what that looks like today and what's
 > hard about it.
 
+> Don't assume that others recognize the problem is one that needs to be solved
 > Is there some concrete issue you cannot accomplish without this?
 > What does it look like to accomplish some set of goals today and how could
 > that be improved?
 > Are there any workarounds that are necessary today?
 > Are there open issues on Github where people would be helped by this?
+> Will the change have performance impacts? Can you quantify them?
 
 > Please focus on explaining the motivation so that if this RFC is not accepted,
 > the motivation could be used to develop alternative solutions. In other words,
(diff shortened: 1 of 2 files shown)


0000-template.md
```diff
@@ -48,9 +76,9 @@ on the impact of the API churn on existing apps, etc.
...
-> This section could also include prior art, that is, how other frameworks in the same domain have solved this problem.
+> This section could also include prior art, that is, how other frameworks in the
+> same domain have solved this problem differently.
 
 ## Unresolved questions
 
-> Optional, but suggested for first drafts. What parts of the design are still
-TBD?
\ No newline at end of file
+> Optional, but suggested for first drafts. What parts of the design are still TBD?
```
lukeed, line 84: I think this is where we need the most prodding, since questions are what reviewers will put forth. AKA, current workarounds (if any) and how they're not enough/acceptable.

benmccann: > current workarounds (if any) and how they're not enough/acceptable

I thought of this as being something to go in the motivation section. I had "What does it look like to accomplish some set of goals today and how could that be improved?", but could explicitly call out workarounds if you think that'd help clarify

lukeed: Ah yea, sorry. Maybe being more direct with "What are the current workarounds, if any" is a better prompt/trigger-word?

benmccann: cool. I just added that

README.md
```diff
@@ -39,4 +41,4 @@ If you are interested in working on the implementation for an 'active' RFC, but
 
 ## Reviewing RFCs
 
-This is where we differ from other teams using this process — we don't have regular core team meetings, and tend to do all our thinking informally, in the open, when time allows.
\ No newline at end of file
+This is where we differ from other teams using this process — we don't have regular core team meetings, and tend to do all our thinking informally, in the open, when time allows.
```
Conduitry, line 44: This should probably be updated, but I'm not sure to what.

0000-template.md
```diff
@@ -10,18 +10,48 @@
...
+> How do different libraries or frameworks implement this feature? We can take
+> design inspiration from others who have done this well and improve upon the
+> designs or make them better fit Svelte.
+
+### Implementation
+
+> Explain the design in enough detail for somebody familiar with the framework to
+understand, and for somebody familiar with the implementation to implement. This
+should get into specifics and corner-cases, and include examples of how the feature
+is used. Any new terminology should be defined here.
```
dummdidumm, line 50: You said you wanted people to not focus on solutions too soon, but this sounds like the opposite to me ("this should get into specifics and corner-cases").

README.md
benmccann: good catch! I took a stab at updating this

0000-template.md
benmccann: yeah, this was something I had trouble with. I tried to encourage a lot of detail before getting to this point (e.g. motivation and background). I think at some point the corner cases have to get answered before final approval though. I just tried to update this section a bit more. Take a look and let me know what you think

dummdidumm: Sounds better! Maybe a final sentence like "you don't need to specify this all up front as many of the implementation details may arise during discussion, but at least put some initial thought into it".

benmccann: I think it would be useful for people to include as much of this as they can, but I don't want it to be a barrier. I updated the wording in this section to make these sound a bit more like suggestions instead of being mandatory. I also updated the PR description to better describe what I was trying to accomplish here

sw-yx:
defining "stages of acceptance" of an RFC could also be helpful to give an idea of forward motion to both the requester and interested parties

antony (approved):
This looks great. I think the final section is particularly useful - basically expect to get out what you are prepared to put in.

antony:
> defining "stages of acceptance" of an RFC could also be helpful to give an idea of forward motion to both the requester and interested parties

Agreed - I wonder if it's always as straight-forward as this right now though, RFCs have many paths to becoming something.

sw-yx:
i mean just off the cuff:

- stage 0: proposed
- stage 1: one maintainer agrees (can be same as proposer)
- stage 2: maintainer consensus achieved
- stage 3: working PR made
- stage 4: PR merged

benmccann:
I wonder if we're that consistent with the ordering of stages. A lot of folks make a working PR first or in parallel to demonstrate that it's possible or what the code would look like. I've also had changes get in without anyone approving the RFC

sw-yx:
just a suggestion, if premature its fine :)

pngwn (approved):

dummdidumm (approved):

PR #34, opened 2020-09-10
Inline components

Rich-Harris:
[**Rendered**](https://github.com/sveltejs/rfcs/blob/inline-components/text/0000-inline-components.md)

### DIFF
--- a/text/0000-inline-components.md
+++ b/text/0000-inline-components.md
@@ -172,5 +172,4 @@ I think this is more confusing than the special element proposal, and doesn't ma
 
 ## Unresolved questions
 
-* Is `<svelte:template name="...">` right? (What if there was a prop called `name`? Should we use `this` instead, or is that confusing?)
 * Can templates be scoped, or must their names be unique within the component? Must they be defined at the top level of the component, or can they be defined anywhere (and inherit context and styles that apply to their subtrees)?
--- a/text/0000-inline-components.md
+++ b/text/0000-inline-components.md
@@ -1,0 +1,176 @@
+- Start Date: 2020-09-09
+- RFC PR: (leave this empty)
+- Svelte Issue: (leave this empty)
+
+# Inline components
+
+## Summary
+
+A new `<svelte:template>` element that defines an inline logic-less component.
+
+## Motivation
+
+In some frameworks, a module can contain multiple components, some of which need not be exported (and are therefore effectively private). In Svelte, there is a strict one-component-per-file rule.
+
+This is nice and simple to understand and explain (not to mention implement), but it does have drawbacks. Users sometimes complain that it causes a proliferation of `.svelte` files, many of which contain very simple components that would be better declared inline.
+
+The alternative is to duplicate markup inline:
+
+```svelte
+{#each images as image}
+  {#if image.link}
+    <a href={image.link}>
+      <figure>
+        <img src={image.src}>
+        {#if image.caption}
+          <figcaption>{image.caption}</figcaption>
+        {/if}
+      </figure>
+    </a>
+  {:else}
+    <figure>
+      <img src={image.src}>
+      {#if image.caption}
+        <figcaption>{image.caption}</figcaption>
+      {/if}
+    </figure>
+  {/if}
+{/each}
+```
+
+Neither is ideal.
+
+## Detailed design
+
+With `<svelte:template>` (🐃), combined with new functionality for `<svelte:component>` in which a string `this` attribute is taken to refer to an inline component, we can avoid both the problems articulated above:
+
+```svelte
+<svelte:template name="image" let:image>
+  <figure>
+    <img src={image.src}>
+    {#if image.caption}
+      <figcaption>{image.caption}</figcaption>
+    {/if}
+  </figure>
+</svelte:template>
+
+{#each images as image}
+  {#if image.link}
+    <a href={image.link}>
+      <svelte:component this="image" {image}>
+    </a>
+  {:else}
+    <svelte:component this="image" {image}>
+  {/if}
+{/each}
+```
+
+Together with the RFCs for [local scoped styles](https://github.com/sveltejs/rfcs/blob/local-styles/text/0000-local-styles.md) and [constants in markup](https://github.com/sveltejs/rfcs/blob/markup-constants/text/0000-markup-constants.md), this gives us everything we need to create self-contained logic-less components, even including slotted content:
+
+```svelte
+<svelte:template name="tooltip" let:dangerousness>
+  {@const danger = dangerousness > 5}
+
+  <div class="tooltip" class:danger>
+    <slot></slot>
+  </div>
+
+  <style>
+    .tooltip {
+      background: white;
+      padding: 1em;
+    }
+
+    .danger {
+      color: red;
+    }
+  </style>
+</svelte:template>
+```
+
+(At the point at which it *does* make sense to turn this into a separate `Tooltip.svelte` component, the extraction is a completely mechanical process that could even be automated by tooling.)
+
+
+### No `<script>` in inline components
+
+It's tempting to suggest that inline components should be allowed their own `<script>` blocks:
+
+```svelte
+<svelte:template name="tooltip" let:dangerousness>
+  <script>
+    $: danger = dangerousness > 5;
+  </script>
+
+  <div class="tooltip" class:danger>
+    <slot></slot>
+  </div>
+
+  <style>
+    .tooltip {
+      background: white;
+      padding: 1em;
+    }
+
+    .danger {
+      color: red;
+    }
+  </style>
+</svelte:template>
+```
+
+I think this would be a mistake: lifecycle and scope would get tricky for component authors to reason about, and (not that this should be an overriding consideration) I would expect it to result in a significant increase in implementation complexity.
+
+
+### Change to `<svelte:component>`
+
+At present, the `this` attribute of a `<svelte:component>` must be either a component constructor or a falsy value. In order to make it work with inline components, we allow `this` to be a string instead. In many cases this could be optimised to be a direct reference to the locally defined constructor, though in the case of a dynamic `this`...
+
+```svelte
+<svelte:component this={string_or_constructor}/>
+```
+
+...it would be necessary to maintain a mapping of names to constructors instead.
+
+## How we teach this
+
+'Inline component' is an obvious term, though it doesn't quite explain `<svelte:template>`. In other template language contexts, the word 'partial' is sometimes used to mean something similar.
+
+Accepting this proposal would mean adding new documentation, and updating the `<svelte:component>` documentation to accommodate `this="string"`.
+
+## Drawbacks
+
+As ever, the concern is that this increases the amount of stuff to learn, and the amount of stuff to implement and maintain.
+
+In a similar vein to ([#33](https://github.com/sveltejs/rfcs/blob/markup-constants/text/0000-markup-constants.md#drawbacks)), it *is* arguably just something that compensates for the lack of power in the template language relative to JavaScript.
+
+## Alternatives
+
+Instead of a new special element, we could use syntax:
+
+```svelte
+{#template "image" { image }}
+  <figure>
+    <img src={image.src}>
+    {#if image.caption}
+      <figcaption>{image.caption}</figcaption>
+    {/if}
+  </figure>
+{/template}
+
+{#each images as image}
+  {#if image.link}
+    <a href={image.link}>
+      {@partial "image" {image}}
+    </a>
+  {:else}
+    {@partial "image" {image}}
+  {/if}
+{/each}
+```
+
+I think this is more confusing than the special element proposal, and doesn't make effective use of existing concepts (`let:`, slots, etc).
+
+## Unresolved questions
+
+* Is `<svelte:template name="...">` right? (What if there was a prop called `name`? Should we use `this` instead, or is that confusing?)
+* Can templates be scoped, or must their names be unique within the component? Must they be defined at the top level of the component, or can they be defined anywhere (and inherit context and styles that apply to their subtrees)?
--- a/text/0000-inline-components.md
+++ b/text/0000-inline-components.md
@@ -1,5 +1,5 @@
 - Start Date: 2020-09-09
-- RFC PR: (leave this empty)
+- RFC PR: ([#34](https://github.com/sveltejs/rfcs/pull/34))
 - Svelte Issue: (leave this empty)
 
 # Inline components
@@ -42,7 +42,7 @@ Neither is ideal.
 
 ## Detailed design
 
-With `<svelte:template>` (🐃), combined with new functionality for `<svelte:component>` in which a string `this` attribute is taken to refer to an inline component, we can avoid both the problems articulated above:
+With `<svelte:template>` (🐃), combined with new functionality for `<svelte:component>` — in which a string `this` attribute is taken to refer to an inline component — we can avoid both the problems articulated above:
 
 ```svelte
 <svelte:template name="image" let:image>


dummdidumm:
I like this a lot. It's also similar to how it works in Angular (`ng-template`) but easier to grasp. My notes:
- Agree that it should not allow for a `<script>`. This is when you really should starting extracting the code into a component.
- About the `name` thing: `this` feels more consistent at first but the semantics are different, so I would say no to this - unless we change the semantics and `<svelte:template this={bla}` means you need a variable `bla` in the `script` which can then be referenced; but I guess this is too verbose. Alternative: We could add special syntax for it, like `#thisIsTheName`. This is how Angular does it. My vote goes to `name` in the end.
- Template scope: They should have access to all variables/methods inside the component, but their style should be independent.
- Top level only or not: I'm split on that one. Top level only would make components more readable (and I guess easier to implement in Svelte core). But I'm not sure if we would lose possibilities. One thing (from a typing perspective) that comes to my mind is stuff like "I'm inside this `#if` so I know that variable `x` is defined and I can safely use it in my template". TypeScript would otherwise error at you "hey this is possibly undefined". But I guess this could be solved by redefining `x` as a `let:y`.
- Speaking of types: How to type `let:y` on the templates?

Conduitry:
I interpreted the question as 'how do we figure out what the types are' not 'what do I do with my keyboard'

dummdidumm:
I meant "how do I add TypeScript typings". But I guess the simple answer for now is "it's type `any`".

lukeed:
I can definitely get behind this.

* Should be defined a top-level only
* Would prefer `this=` for both `<svelte:component>` and `<svelte:template>`
* inherits **everything** because it was _defined_ in that context:
    * local `<style>` overrides styles
    * no `<script>` to avoid complexity
    * use `{#const}` to define local var(s)

A possible alternative to `<svelte:component this="image"/>` would be something like this:

```svelte
<svelte:template this="Image" let:src>...</svelte:template>

<Image {src} />
```

If we go with that route, it _could_ make sense for `Image` to be fully isolated (inherits nothing) since it looks like any other component. This requires that the name (`"Image"`) be capitalized as a linter/compiler rule.

Another option is to enforce the name be lowercased (eg, `"image"`), and is used as such:

```svelte
<svelte:template this="image" let:src>...</svelte:template>
<image {src}/>
```

An obvious issue here with the lowercased approach is HTML/CE element conflicts.
A secondary issue (to both) is that it might be pretty tricky to get ESLint/TS tooling to figure out `Image` came from `this="Image"` -- but I've no idea.

halfnelson:
I had the need for something like this in Svelte Native.
The nativescript ListView control takes a "template" as a slot and instantiates it. I went with something based on this
https://svelte.dev/repl/d8c056f37ce949d7a7d1be5afe9cac53?version=3.25.0

```html
<script>
	import AsComponent from './AsComponent.svelte'
	let menuItem
</script>

<AsComponent bind:component="{menuItem}" let:props={{val}}>
    <h3>This could be a  menu item {val}</h3>
</AsComponent>

<svelte:component this={menuItem} val="1" />
<svelte:component this={menuItem} val="2" />
<svelte:component this={menuItem} val="3" />
<svelte:component this={menuItem} val="4" />
```

Not ideal, but there was a nice advantage to having the component be a full svelte component. It allowed me to pass instances around and instantiate them outside of the component or pass them into children as props (although we have slots for this)

rob-balfre:
I like it.

Side note... Would this allow tree shaking for parts of a component not in use?  For example svelte-select has loads of features (too many TBH!) and I'd love to allow parts of it to be excluded from the compiled output if not in use. So if parts of svelte-select were rewritten as inline components and linked to props and those props are not used then rollup (no example) would not compile and output that code?

knobo:
What happens if you alter an input value in an inline template
``` svelte
<svelte:template this="Image" let:src>
      <img on:click={() => src="/img/clicked.png"} {src} />
</svelte:template> 
```
Is it a good idea or not?

Are there done any considerations on if input properties should be const?
And how would dispatch work, if there no inline scripts?
Could it be passed in as a prop somehow?
``` svelte
<svelte:template this="Image" const:src  dispatch>
      <img on:click={() => dispatch({detail:"whatever"})} {src} />
</svelte:template> 
```

dummdidumm:
I would say the answer to these questions is "the template shares the scope with the parent, so it can both read and write values defined in `<script>`".

halfnelson:
So a template is just like an each block, except all instances don't have to be in the one spot

PaulMaly:
I think in general this proposal makes sense, but I'm not sure it's the best way to make it.  Few questions & remarks:

- I don't think the inline component (partial / template) should have its own styles and state. I believe it should work completely in the context of the SFC.
- I think we can just `<template>` tag for this purpose:

```svelte
<template name="image" let:image>
  <figure>
    <img src={image.src}>
    {#if image.caption}
      <figcaption>{image.caption}</figcaption>
    {/if}
  </figure>
</template>
```
Seems we already use `<slot>` tag in a specific svelte-way.

- I'm not sure we should change the behavior of `<svelte:component>`, it's confusing me a little. Maybe we could jointly consider another proposal [#2324 ](https://github.com/sveltejs/svelte/issues/2324) which is very popular and already has a PR. And just extend `<svelte:element />` idea and give the ability to pass `template` element to that tag with more wide capabilities.

```svelte
<script>
   let partialName = 'image';
</script>

<template name={partialName} let:image>
  <figure>
    <img src={image.src}>
    {#if image.caption}
      <figcaption>{image.caption}</figcaption>
    {/if}
  </figure>
</template>

{#each images as image}
  {#if image.link}
    <a href={image.link}>
      <svelte:element this={partialName} {image} />
    </a>
  {:else}
     <svelte:element this={partialName} {image} />
  {/if}
{/each}
```
- I think we should avoid any context separation of the inline component and the component which contains it. It would be better if content inside `<template>` tag would be referenced to state & styles almost the same as `<slot>` tag.
- I think `inline component` is not the best term for this and confusing. I prefer `partial template` - but you want to know so here we go.

AlexxNB:
I prefer an alternative version in the RFC(partialy). In my opinion, inline-component declaration should be in control block(like each, if, await and others), but its using in the template should be as element. I [suggested](https://github.com/sveltejs/rfcs/pull/18#issuecomment-623008985) similar syntax for slots also in the past.

```html
{#template:Button text,type}
    <button class:primary={type=='primary'}>{text}</button>
{/template}

...

<Button text="Click me" type="primary />
<!-- OR -->
<svelte:element this="Button"  text="Click me" type="primary />
<!-- OR -->
<template:Button   text="Click me" type="primary />
```

PierBover:
To be honest, I think I'm one of the people that motivated this feature by complaining on Twitter and Github. :)

IMO this is going in the right direction. Obviously we will never have the flexibility of JSX or HS with a SFC format, and that's totally fine. Logic-less pieces of templates are a fine solution and would solve 80% of the problem.

Although I think it would be nice to refine and simplify this even further. I agree with @PaulMaly here:

> I think inline component is not the best term for this and confusing. I prefer partial template - but you want to know so here we go.

Since these inline components do not have their own logic, I don't think these can be called components at all. We'd still need to write event handlers and whatnot in the `<script>` of the file. Wouldn't it be more consistent to do the same with styles and just call these template partials or fragments?

It would be easier to understand and explain too. Instead of thinking about it like a component but with exceptions, it's just a piece of template that can be reused in the file.

PaulMaly:
@AlexxNB Just to be sure that we all know how it was in Ractive - [In-template partials](https://ractive.js.org/api/#in-template-partials)

```html
{{#partial image}}
  <figure>
    <img src="{{src}}">
    {{#if caption}}
      <figcaption>{{caption}}</figcaption>
    {{/if}}
  </figure>
{{/partial}}

{{#each images}}
  {{#if link}}
    <a href="{{link}}">
      {{> image }}
    </a>
  {{else}}
     {{> image }}
  {{/if}}
{{/each}}
```

stephane-vanraes:
> IMO this is going in the right direction. Obviously we will never have the flexibility of JSX or HS with a SFC format, and that's totally fine. Logic-less pieces of templates are a fine solution and would solve 80% of the problem.

I would say the exact opposite: this is a move in the wrong direction.

In JSX I often find issue with opening a component (say _House_) and having to dig through all kind of code related to _Room_ and _Roof_ before getting to the actual code for the House I am looking for.  Even if those components are not used elsewhere, to me it doesn't make sense to have them inlined at all.  It also feels like a violation of the Single Responsibility principle: my House.svelte should only concern how to render a house, not how the individual rooms are dealt with.

On top of that, how does a solution like this handle testing ?
In JSX it's easy to test the 'inline templates': just export them as well ! (completely undoing the reason why you inlined them in the first place)
In order to make these inline templates testable (and one day people will ask for it) a method of exposing them will have to established as well, this in turn will open the door to having files simply filled with _templates_ that are exported in real components.

That said, having a lot of files can be a problem.
But the problem does not lay in file management (that can be easily solved by using good naming, folder structures, etc.
For instance the following structure still allows me to simple do `import House from './House`:

```
- /House
      - index.svelte
      - Room.svelte
      - Roof.svelte
``` 

The problem with having lots of 'dumb' files is the extra overhead created from having so many components in the compiled code, bringing the code closer to the inflection point where the generated bundle is larger than alternative frameworks.

PierBover:
> In order to make these inline templates testable (and one day people will ask for it) a method of exposing them will have to established as well, this in turn will open the door to having files simply filled with templates that are exported in real components.

If you consider these as template partials instead of inline components then these are just part of the same component. It would follows the same model as one component per file, SRP, etc.

When testing React you don't really test a piece of JSX in a variable inside a component.

pngwn:
Yeah. We don't _want_ people testing inline components as they are implementation details. And we don't not want to encourage the testing of implementation details.

pngwn:
I think Svelte can only go so far when encouraging good practices, like any other library. For example, I consider deep folder structures an anti-pattern than negatively impacts discoverability and makes onboarding very challenging.

I think inline components solve certain issues and could add a degree of flexibility that we don't currently have. People could certainly over-use them but people also currently write incredibly large components that are difficult to understand too, it could be argued that Svelte encourages _that_ as well.

I think some of the reason for this centre around the cost of additional components, and confusion over how Svelte scales. In most cases these are misconceptions but there is some truth in them. We could do better when it comes to highlighting that scaling is not much of an issue in practice but I also think additional flexibility in certain cases is valuable to users. We aren't throwing all of our principles away here, just suggesting the reuse of inline fragments.

@Rich-Harris Building on a conversation we had a while ago about fragments, what about using `svelte:fragment` instead:

```svelte
<svelte:template name="image" let:image>
  <figure>
    <img src={image.src}>
    {#if image.caption}
      <figcaption>{image.caption}</figcaption>
    {/if}
  </figure>
</svelte:template>

{#each images as image}
  {#if image.link}
    <a href={image.link}>
      <svelte:fragment this="image" {image}>
    </a>
  {:else}
    <svelte:fragment this="image" {image}>
  {/if}
{/each}
```

arxpoetica:
+1 for "partial." People know this nomenclature vs (confusing) "template."

dummdidumm:
Funny, for me it's the opposite. Having an Angular background, `template` feels natural to me.

pngwn:
For reference [`svelte:fragment`](https://github.com/sveltejs/svelte/pull/4556) if I recall this was one of the reasons for the `fragment` rename, so we could use it in other places (this being one of the suggestions). So it wouldn't be new, although that PR is still open, there is broad agreement that the PR should be merged at some point.

pngwn:
@benmccann That won't work because a partial can have slotted content.

pngwn:
Thinking about it more is the inline component the `fragment` in this case? Maybe that makes more sense.

PaulMaly:
> For reference svelte:fragment.

Maybe we can use `<svelte:fragment>` in different ways based on its attributes, eg:

```svelte
<svelte:fragment slot="foo" />
<svelte:fragment template="bar" />
```

> That won't work because a partial can have slotted content.
@pngwn Btw, I'm not sure it should.

halfnelson:
I like the concept of `svelte:fragment` being a collection of reactive dom nodes.
Introducing `svelte:fragment` could solve one of the problems with slots: you need a dom node to put the `slot=` attribute.

Considering we already use the [slot](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot) element for slots, using the [template](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template) makes sense too:

> The HTML Content Template (`<template>`) element is a mechanism for holding HTML that is not to be rendered immediately when a page is loaded but may be instantiated subsequently during runtime using JavaScript.

We could do things like

```html
<template name="image" let:image>
  <figure>
    <img src={image.src}>
   </figure>
</template>

<SomeComponent>
  <svelte:fragment slot="aside" template="image" />
</SomeComponent>
```

The advantages of this are: 
 * only one new special tag `svelte:fragment`
 * similar behaviour to slots (where the attribute name refers to the element type and the value refers to the name) (`slot=<slotElement.name>` `template=<templateElement.name>`)

lukeed:
I don't think we can do `template` (no namespace) because that's already a valid tag. It'd be unique in that Svelte completely mangles its native behavior and/or removes it from the component tree.

PaulMaly:
I like how @halfnelson summarised the best ideas.

@lukeed How about `slot` and `style` tags? They also a valid tags whose native behavior completely mangles by Svelte.

thgh:
When thinking of inline components, I expected to be able to export two components from one SFC. E.g. `import Card, { Carousel } from ...` . Where Carousel is a custom carousel component optimized for that specific Card component. So I thought it would make sense to see this as a virtual file and allow a script in inline components. (With full encapsulation)

If it's not a full component, like a partial or fragment, I would not expect style and var encapsulation. Would that be powerful enough?

PaulMaly:
@thgh actually, you don’t need any kind of “inline components” to import multiple components from the single point. In Svelte you can just create a folder for component(s) with index.js and re-export all thing from there. It’s not a big deal and shouldn’t be a reason  to add new features to the framework.

dummdidumm:
Things like "how to test this?" or "can I export this as well?" are _not_ what this RFC is about IMO. It's about saving some boilerplate when doing something within the component. That template/fragment/partial/whatever-we-will-call-it is _private_ to the component.
(We should definitely not call it "inline component" so people don't get the impression this is a full-fledged inner component)

stephane-vanraes:
> Things like "how to test this?" or "can I export this as well?" are not what this RFC is about IMO

That is perfectly true, I brought up exporting and testing up now, because I expected it to be requested sooner or later (as it stands, later seems to be the next day)

> It's about saving some boilerplate when doing something within the component. That template/fragment/partial/whatever-we-will-call-it is private to the component.

This is in fact an excellent idea, as was mentioned before it does solve a specific pain that developers currently have and are requesting.

> (We should definitely not call it "inline component" so people don't get the impression this is a full-fledged inner component)

I do take issue as well calling it an "inline component", not an expert in naming things myself, but another name might indeed be better to make clear these are not _components_, and do not have their own style, logic, ... and should only be used within the context of this SFC.  This might be slightly harder to "teach" than originally expected though.

For naming, what about 
- partial component (this might reflect that it's _part of_ a component ? but might be expected to be a separate file.
- inline fragment (I think this covers them fairly good, as most people consider fragments to be kind of dumb, markup only things)

aidarkhanov:
@stephane-vanraes Good point regarding the naming; “component” implies to have logic and internal state combined together. “Fragment” is a better name for denoting a smaller dependent part of a bigger piece, without internal state.

To clarify, a component is an independent and complex unit of your UI, big or small.
Whereas a fragment can be considered a smaller part of a component that only fits well in a certain context.

Fragments by design should not be involved in a patchwork, tailoring is a better word for fragments.

lukaszpolowczyk:
@AlexxNB
Oh yes! `{#template:Name}{/template}` is something I wanted to suggest. Earlier, I [suggested](https://github.com/sveltejs/svelte/issues/4478) `<template:Name></template>` to declare inline component.

@AlexxNB @PaulMaly
But why can't this be used with a plain `<Name />`?
Why does it have to be `<svelte:component this="Name" />` or `<svelte:element this="Name" />`?

The combination of `#template:Name` and plain `<Name />` gives the simplest, cleanest syntax.

PaulMaly:
@lukaszpolowczyk Actually, not just a name there. `<Name />` can be in cases available for static analysis. Any dynamic things occur in runtime, so it should be some JS runtime stuff, like expressions or class name, eg.

```
<svelte:component this={ClassName}  />
```

lukaszpolowczyk:
@PaulMaly I do not understand what you mean. Can you write in a different way?

RoystonS:
I'm worried about the lack of ability to provide type information for the `let` props in the current proposal.

In React, I can write a quick typed inline component:
```
interface ImageProps {
  image: { src: string; caption: string; };
}
const Image = (props: ImageProps) => <figure>...
```
I then get compile-time type-checking and IDE Intellisense to ensure that when I use the `<Image>` component I haven't accidentally passed in a string URL to `image` instead of the required object.

With the Svelte example in the RFC
```
<svelte:template name="image" let:image>
```
there's no ability to provide type information for the `image` property. Without syntactic help, it would forever be stuck at `any`, which removes the helpful safety net that TypeScript provides.

Being able to write `<svelte:template name="image" let:image:SomeType>`  would be better. Not terribly natural, and wouldn't be able to support complex anonymous inline types such as `:{src: string; caption: string}` but being restricted to named types would be better than being forced to `any`.

lukaszpolowczyk:
@RoystonS Then the syntax `{#template:Name prop1:string, prop2:number} {/template}` or `{#template "Name" prop1:string, prop2:number} {/template}` would be better.

RoystonS:
@lukaszpolowczyk Potentially. That _could_ in theory support more complex types such as
`{#template:Name image:{ src: string; caption: string}}` (like my React example) because it's not space-separated. BUT, doing so would require a more complex parser which would be at the mercy of new pieces of TypeScript syntax, which wouldn't be good.

I'd suggest that merely accepting simple named types might be enough. Even for a generic property type
`{#template:Name prop1:Something<number,string>}`, that could always be rewritten
```
type SpecificSomething = Something<number, string>;
{#template:Name prop1:SpecificSomething}
```
Not as beautiful, but way simpler for Svelte to maintain, and, most importantly, it allows keeping type safety.

So, if sticking with simple type names, adding type information to a prop doesn't make a particular case for `<svelte:template name="foo" let:prop:Type>` versus `{#template:foo prop:Type}` versus `{#template "foo" prop:Type}`.  I do _have_ a personal preference out of those, but that's a different matter.

antony:
I'm going to vote for having `fragment` as opposed to `template` because `template` will make it appear that all html should be wrapped inside it and that's just horrible.

Fragment correctly indicates that this is a fragment of code which is a sub-component.

And as a final remark I'd just like to say that I still don't support this RFC at all!

arxpoetica:
> I still don't support this RFC at all

@antony reason? Are you opposed to the idea of partials in general? They're definitely something useful in other frameworks.

lukaszpolowczyk:
Do module blocks make a difference?
https://twitter.com/DasSurma/status/1353831156098150400

intrnl:
Chiming in because this deserves the attention that it needs.

Fragment seems like a fitting name for this, and currently this already solves the issue that people has, which is being able to reuse parts of a template within a component.

Personally It might be worth allowing the usage of `<script>` within fragments, and perhaps even allowing the exporting of fragments. We don't have to recommend using `<script>` inside fragments, especially if one's needs is to have logicless templates, but as for why I wanted the two things I mentioned...

One of the things I've grown accustomed to doing in React is breaking up components into small "primitives" that can be composed together by other components, which has allowed me the flexibility to mix and match parts without any worries. This doesn't seem all that convenient on Svelte currently given that I would have to create a new file for each primitives that I make, so being able to export fragments would be a nice thing to have.

benmccann:
Here's a pre-processor that supports multiple components in a single file: https://github.com/srmullen/svelte-subcomponent-preprocessor

madeleineostoja:
I still think this RFC has a lot of validity that that preprocessor doesn't address, namely having a 'blessed' way to create reusable chunks of markup in Svelte itself (which is an extremely common use-case), and the strictness of this just being stateless template rather than a full blown svelte SFC inside another SFC. Writing those inline components to disk as the above preprocessor does feels like a hacky bandaid in comparison.

It's a pity momentum on this proposal died down, is anyone on the core team opposed to it , or is there just limited interest/time to move it forward?

ryanatkn:
I'm not adding anything new here, but I want to elaborate the pro-RFC case for people who don't like it.

Currently, creating reusable markup requires creating a new file, exposing it as app-wide API. Projects have files that aren't intended to be used independently, breaking encapsulation and bloating the source tree. Developers are unable to hide API and complexity within desired bounds.

Consider writing JS where every top-level helper function, even the 1-liners, could be imported by the entire app, and nothing is local to a module. It would lower the expressivity of the language, limiting the developer's ability to communicate and enforce how code should be used. From the intra-app user's perspective, it's more complicated because the app namespace has things that would otherwise be hidden. What should be private code may later be used in a way that wasn't originally intended, tangling concerns and complicating future change.

I think the semantics of calling the feature inline _components_ makes this muddier than it should be, hence the support for alternative names like fragment and template. The RFC changes only part of Svelte's simple "one component per file" design -- there will still be one local script tag and one component export, so it's still one component per file in that sense. The proposal retains the simple public component interface while extending components with private abstraction powers, of the sort we expect from other languages.

madeleineostoja:
Wholeheartedly agree with everything @ryanatkn said 👌🏼

madeleineostoja:
Also, without having read every comment in the thread yet, I’d like to speak highly in favour of the declarative `<svelte:template>` over curly braces syntax for this feature. Again I think the best mental model for this is a Svelte-like version of the native `<template>` tag. Whereas currently Svelte’s curly brace API is only for control flow in markup, which I think would be muddied by adding a template definition.

arxpoetica:
Oof. Sorry. Vote NO to `<svelte:template>`. For someone coming from Vue, the mental model will land wrong, since that's the top level wrapper. This is true of many DSLs actually.

`partial` or `inline` has a more familiar precedence: https://handlebarsjs.com/guide/partials.html

madeleineostoja:
I think just because Vue hijacked template as their top level wrapper doesn’t mean svelte shouldn’t use it for a different purpose, the purpose of the native template element is exactly this RFC

arxpoetica:
Hadn't known about actual `<template>` tag (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template).

I still hesitate with that, however. I wouldn't say "the purpose of the native template element is exactly this RFC," but rather, "this RFC models similar behavior."

If people feel that this is the only place where we'd model similar behavior as `<template>`, so be it, but I'd hesitate to adopt that namespace where it _**might**_ come up in the future under a different (more appropriate, Vue-like) use case. Template seems like a higher-level component type thing.

I'm still partial (pun) to `partial` or `inline`, but I'll defer to the majority (or other maintainers).

antony:
> > I still don't support this RFC at all
> 
> @antony reason? Are you opposed to the idea of partials in general? They're definitely something useful in other frameworks.

I'm less averse to it after seeing the proposed implementation. I don't like inconsistency within apps, i.e. not knowing if a component is in a file or embedded, for example, but `<svelte:component>` is pretty reasonable.

PuruVJ:
How would we define types for the props with the `let:` syntax? Asking for the TypeScript users

lukaszpolowczyk:
This should be possible with inline components:
```svelte
<!-- Component.svelte -->
<svelte:template name="Component2">
   <span>c2</span>
</svelte:template>
c1
<slot Component2={Component2} />
c1

<!-- App.svelte -->
<Component  let:Component2={Component2}>
   <div>
      s1
      <svelte:component this={Component2}/>
      s1
   </div>
</Component>
```

I checked that regular components can be passed through the slot parameter.

I know there is resistance to using let `export` inline component, but here it is only releasing in the scope of the called component `Component`, and actually "returns" `Component2` to the inside via the slot.
You could call it an "Inverted Slot".

I think there should be no doubt about that?

Florian-Schoenherr:
Isn't the "problem" of small components solved by opening 2 editor windows of whatever IDE?
Or is there a tradeoff in performance if you have 2 components instead of one? (which would not be solved by this I just noticed, because it's still 2..n components, just in 1 file)

intrnl:
> Isn't the "problem" of small components solved by opening 2 editor windows of whatever IDE? Or is there a tradeoff in performance if you have 2 components instead of one? (which would not be solved by this I just noticed, because it's still 2..n components, just in 1 file)

It's because they're small that it's unpractical to have them in a separate file.

Yes, you can open multiple editors, but you have to keep reminding yourself that you have to do this when you want to mess around with it.

Florian-Schoenherr:
@intrnl You're right. I would just like to keep software simple.
I think more people will just make really big files... though the REPL would be easier too, I guess.

aradalvand:
@Florian-Schoenherr This is also particularly useful when you want to move a part of your component around based on some condition. Having to create a separate `.svelte` file in cases like this would be a bummer. Plus you'll also have to declare and pass down props and sometimes do a whole lot more. The inline components feature eliminates the need for all of this.

Florian-Schoenherr:
@AradAral good one. From a component lib author perspective, I would go for an easier way to pass down stuff, see the different ideas in rfcs and issues around a `forward:` directive, or `on:*` or spread binding.

kevmodrome:
We discussed this RFC on Svelte Radio today (not released yet) and I'm getting more and more skeptical towards this. This is very much a niche use-case in the grand scheme of things and it definitely ups the complexity level.

PierBover:
I have to admit I feel less strongly about this as I did when I started using Svelte two years ago.

Objectively, I still think JSX is more flexible and there are definite advantages to this flexibility like being able to create multiple components in single file.

But OTOH I now also think that adding multiple components in a single `.svelte` file would be quite inelegant, regardless of the implementation. I've come to appreciate that Svelte's simplicity and elegance are really its biggest assets (less code, less abstractions, less cognitive load, etc).

IMHO Svelte should trend towards an even simpler and more elegant workflow. Just like we saw with the recent inclusion of the [style properties](https://svelte.dev/docs#template-syntax-element-directives-style-property) which will save us a lot of mangling with strings for dynamic styles.

lukaszpolowczyk:
@kevmodrome @PierBover It pains me that someone can write like that at all.

I understand that it is expensive to maintain the "extra" syntax - if that **is the ONLY reason**, then ok.

But I will never agree with the statement that inline component or similar contructions are unnecessary or somehow hurt.

See if inline component is not analogous to e.g. `@const`?
After all, inline component is just **making you not have to write the same SHORT markup twice**.


**In general** I have plans to do a review of the flow of things in svelte, type:
```
svelte-script -> svelte-script
svelte-script -> svelte-markup
svelte-script -> svelte-style
svelte-markup -> svelte-script
svelte-markup -> svelte-markup
svelte-markup -> svelte-style
svelte-style -> svelte-script
svelte-style -> svelte-markup
svelte-style -> svelte-style
```

etc., - communication of everything with everything.
It is about passing simple data, as well as whole blocks of "body" of a given type(`script`, `markup`, `style`).

E.g. there are also some shortcomings in communication with style. But I will not write everything here.

**In this thread** we have the issue of using the markup "body", that is the communication `svelte-markup -> svelte-markup`.

In fact, an inline component is equivalent to a function from `svelte-script` - with `svelte-script` the body is the code in the function, with svelte-markup the body is the markup block.

Of course I'm not saying that every construct from script(and every other) has to be moved to markup and style etc.
But **I'm certainly saying that there should be constructs for communication/transfer/use of _body/simple data_ between `script`-`markup`-`style`.**

stephane-vanraes:
This is a feature that is often enough requested to at least be considered.

That said I am not a proponent either as I am very much in favour of the 'Single File Component' principle Svelte has had up to now.

Since the comparison with JSX came up, I will also say that this is the very first thing I actually advice clients when using React: remove inline components and stick to one component per file,  the result has always been the same: easier to read, maintain and debug code.  This with the added benefit that in JSX it is clear where one inline component starts and where it ends because of the function signature, something that _might_ be less clear with Svelte.

All in all, with the API surface of Svelte growing it means conflicting views on how to use it will emerge, with this comes that eventually teams will have to agree internally on coding practices to ensure consistency of their code base.  

This means some teams might allow inline components (not mine) and others will not (mine), just like usage of features in other frameworks can be limited by an organization.

But this is a **good** thing, it shows Svelte is flexible and mature, able to change to different environments and practices.

aradalvand:
> We discussed this RFC on Svelte Radio today (not released yet) and I'm getting more and more skeptical towards this. 

@kevmodrome I strongly disagree. There are perfectly legitimate use cases for this, cases where creating a separate component would be too much, such as the one I mentioned in [this comment](https://github.com/sveltejs/rfcs/pull/34#issuecomment-1007161818). And there are use cases as well, for instance, when you have a piece of markup that's specific to a single component (so it doesn't make sense to extract it into a separate component) but it's repeated more than once in your component.
I totally agree with @lukaszpolowczyk:

> See if inline component is not analogous to e.g. @const?
After all, inline component is just making you not have to write the same SHORT markup twice.

I don't understand the reason for you getting more skeptical about this?! It doesn't add much complexity, there are legitimate use cases for it, and people who don't want to use it will not have to.

ponderingexistence:
Personally I have actually encountered more use cases for this than the `{@const ...}` thing.
I am very much in favor of it. I don't think it has any major cons (apart from some extra complexity) but it enables very useful scenarios.

madeleineostoja:
I'd agree that multi-component files go against the philosophy of Svelte, and I think they could encourage problematic edge cases especially since Svelte templates don't have a single root node.

However I do think that defining reusable snippets of markup — whatever you want to call it (templates, consts, fragments) — has a lot of value and doesn't muddy the waters. You're still living in the same component instance, all you're doing is DRYing out markup, which feels very much in line with Svelte's philosophy. If one doesn't already exist I think a separate RFC should be opened for that specific proposal.

aradalvand:
> However I do think that defining reusable snippets of markup — whatever you want to call it (templates, consts, fragments) — has a lot of value and doesn't muddy the waters. You're still living in the same component instance, all you're doing is DRYing out markup, which feels very much in line with Svelte's philosophy.

Exactly; well said.

> If one doesn't already exist I think a separate RFC should be opened for that specific proposal.

I think this current RFC is essentially the same concept, the name is what's a little misleading here, because these aren't really "inline components" but rather, as you said, **component-level reusable bits of markup**.

I'd suggest we call this something else — like "templates", which also matches the `<svelte:template>` tag, or "partials" — but again, the underlying construct is perfectly legitimate.

aradalvand:
Okay wait a second, one thing I'm struggling to understand even after reading the [proposal](https://github.com/sveltejs/rfcs/blob/inline-components/text/0000-inline-components.md) though is whether or not an "inline component" (or template or whatever) would have direct access to the variables declared in its containing component?
I was assuming it would, of course; but I'm now confused. Would the following work, for example?
```svelte
<script>
    let someVariable = 'Hello, World!';
</script>

<svelte:template name="foo">
    {someVariable}
</svelte:template>

<svelte:component this="foo" />
```

If not, then there would be little benefit to this approach, over separate components, and I would probably have to create another proposal that includes this capability as I don't think is really what I was talking about. But if anybody knows the answer to my question, please let me know.

kevmodrome:
> @kevmodrome @PierBover It pains me that someone can write like that at all.
> 
> I understand that it is expensive to maintain the "extra" syntax - if that **is the ONLY reason**, then ok.
> 
> But I will never agree with the statement that inline component or similar contructions are unnecessary or somehow hurt.
> 
> See if inline component is not analogous to e.g. `@const`? After all, inline component is just **making you not have to write the same SHORT markup twice**.
> 
> **In general** I have plans to do a review of the flow of things in svelte, type:
> 
> ```
> svelte-script -> svelte-script
> svelte-script -> svelte-markup
> svelte-script -> svelte-style
> svelte-markup -> svelte-script
> svelte-markup -> svelte-markup
> svelte-markup -> svelte-style
> svelte-style -> svelte-script
> svelte-style -> svelte-markup
> svelte-style -> svelte-style
> ```
> 
> etc., - communication of everything with everything. It is about passing simple data, as well as whole blocks of "body" of a given type(`script`, `markup`, `style`).
> 
> E.g. there are also some shortcomings in communication with style. But I will not write everything here.
> 
> **In this thread** we have the issue of using the markup "body", that is the communication `svelte-markup -> svelte-markup`.
> 
> In fact, an inline component is equivalent to a function from `svelte-script` - with `svelte-script` the body is the code in the function, with svelte-markup the body is the markup block.
> 
> Of course I'm not saying that every construct from script(and every other) has to be moved to markup and style etc. But **I'm certainly saying that there should be constructs for communication/transfer/use of _body/simple data_ between `script`-`markup`-`style`.**

Sorry I don't understand this comment at all, could you re-phrase it somehow?

> > We discussed this RFC on Svelte Radio today (not released yet) and I'm getting more and more skeptical towards this.
> 
> @kevmodrome I strongly disagree. There are perfectly legitimate use cases for this, cases where creating a separate component would be too much, such as the one I mentioned in [this comment](https://github.com/sveltejs/rfcs/pull/34#issuecomment-1007161818). And there are use cases as well, for instance, when you have a piece of markup that's specific to a single component (so it doesn't make sense to extract it into a separate component) but it's repeated more than once in your component. I totally agree with @lukaszpolowczyk:
> 
> > See if inline component is not analogous to e.g. @const?
> > After all, inline component is just making you not have to write the same SHORT markup twice.
> 

I have never said that there aren't use-cases, I'm saying they are **niche** use cases. There aren't enough of them to justify the complexity of this. And to be clear, when I'm talking about complexity I'm not only talking about the internals - I'm worried we'll see a departure from easily digestible, understandable code. 

> I don't understand the reason for you getting more skeptical about this?! It doesn't add much complexity, there are legitimate use cases for it, and people who don't want to use it will not have to.

This is not a good argument. Much like Svelte is opinionated on the styling issue that I know you're very passionate about, it can be opinionated on other things as well. 

> I'd agree that multi-component files go against the philosophy of Svelte, and I think they could encourage problematic edge cases especially since Svelte templates don't have a single root node.
> 
> However I do think that defining reusable snippets of markup — whatever you want to call it (templates, consts, fragments) — has a lot of value and doesn't muddy the waters. You're still living in the same component instance, all you're doing is DRYing out markup, which feels very much in line with Svelte's philosophy. If one doesn't already exist I think a separate RFC should be opened for that specific proposal.

@AradAral @madeleineostoja I'd love to get some real concrete use-cases for this as well as the frequency at which you encounter them. In those cases, would creating a new file really be that much work? From the way you argue about this it sounds like it's something you're running into all day and all night. I think I've had to deal with this a handful of times over the course of 3 years.

lukaszpolowczyk:
> Sorry I don't understand this comment at all, could you re-phrase it somehow?

@kevmodrome Sorry, I can't. (ⴲ﹏ⴲ)/

---

1. If the only reason to abandon inline components is the cost of maintaining Svelte, then I understand the problem.
The other reasons for abandoning inline components I don't understand. I recognize that the other reasons for abandoning inline components are false.

---

2. `@const` is similar to inline components.
Thanks to it, you don't have to write a piece of code many times.
If we have entered `@const`, all the more reason to enter inline components.

---

3. Svelte is divided into:
  
  * `svelte-script` - code
  * `svelte-markup` - tags like div, span etc.
  * `svelte-style` - css
  
  The inline components syntax for `svelte-markup`, is equivalent to the `function` syntax for `svelte-script`, e.g. inline component:
  
  ```svelte
  <svelte:template name="tooltip" let:dangerousness>
    {@const danger = dangerousness > 5}
  
    <div class="tooltip" class:danger>
      <slot></slot>
    </div>
  </svelte:template>
  ```
  
  is equivalent to, for example. `function`:
  
  ```svelte
  function name () {
    /* code */
  }
  ```
  
  That is:
  
  * You create a `function` and use in the second and third place.
  * You create an inline component and use it in the second and third place.

---

4. Both `function` and inline component are very basic syntax:
* What is between `<svelte:template name="tooltip" let:dangerousness>`
and `</svelte:template>`, is `svelte-markup-body` (that's what I called it, you can give a better name).
* What is between `function name () {` and `}
`, is `svelte-script-body`.

I'm not saying to create equivalents of every syntax from `svelte-script` to `svelte-markup` etc, but the ability to reuse `svelte-markup-body` as well as `svelte-script-body` is abosolutely basic syntax.
 
---

5. I also wrote that it would be useful to do a comprehensive review of reuse, syntax portability, and communication between `svelte-script`, `svelte-markup`, and `svelte-style`.
But that's a broader topic, I just mentioned it.

---

I don't know if you understand more now? (ⴲ﹏ⴲ)/

kevmodrome:
> 1. If the only reason to abandon inline components is the cost of maintaining Svelte, then I understand the problem.
>    The other reasons for abandoning inline components I don't understand. I recognize that the other reasons for abandoning inline components are false.

It's not the only reason. Complexity in code bases, "glancability" and separation of concerns will all suffer if this was implemented.

> 2. `@const` is similar to inline components.
>    Thanks to it, you don't have to write a piece of code many times.
>    If we have entered `@const`, all the more reason to enter inline components.

I don't think this is an argument to include it at all. The feature should be able to stand on its own.

> 3. Svelte is divided into:
> ...

I think I see where you're coming from here, though I don't find the argument particularly convincing. Besides, you can already re-use svelte-markup-body, it's called `make a new component`. One of the worst parts of React is the ability to stick an arbitrary number of components (functions, I suppose) in one file. I would very much like to avoid that in Svelte.

I don't think anyone has sufficiently showed that there are enough situations where this feature is needed to warrant implementing it. Especially with regards to the downsides already mentioned.

PierBover:
Just to be clear, I'm not opposed to implementing inline templates or fragments.

I don't like having duplicated markup but I've found this is not a common issue for me (once I tamed my DRY obsession from JSX). It's not even close to being one of my biggest pain points when using Svelte.

> I think this current RFC is essentially the same concept, the name is what's a little misleading here, because these aren't really "inline components" but rather, as you said, **component-level reusable bits of markup**.

@AradAral but in that case, wouldn't it make more sense to solve this in a preprocessor instead? Or even include it [svelte-preprocess](https://github.com/sveltejs/svelte-preprocess)?

ryanatkn:
I agree it's a fairly niche need. At the same time, many developers work in niches. The energy behind the PR suggests that it addresses a real pain point, more than a theoretical or aesthetic preference, but 1) we're stuck with anecdotes, 2) every example has its workarounds, and 3) how much pain is it to how many people?

I subjectively rate it as desirable probably higher than most people, but I'd also understand if the RFC was rejected. The feature adds significant complexity and costs, and breaks expectations that some people find important. Is it worth the pain relief? That question doesn't have one answer.

madeleineostoja:
I would also agree that this isn't a huge pain point, but a nice-to-have. At the end of the day we're talking about convenience and cleanliness here, it's not like having reusable bits of markup would make something previously impossible suddenly possible, it's just an ergonomics thing. But a huge part of Svelte in DX, so I do think it's worth considering, even if it's not a high priority in Svelte's roadmap. 

I don't have examples on hand, I've probably run into situations I would have reached for this feature maybe half a dozen times in my own work (typically CMS-backed websites where I would have previously used NextJS) over the last 6 months as well. Basically wherever I would have previously thrown a chunk of JSX in a const. Again none of these cases were a dealbreaker just having to repeat some markup, it would've just been nice.

Given that (for me at least) it's a DX thing, I would reiterate that the complexity and unforseen edge cases of having legitimate inline components would not be worth it right now. I'm just talking about bits of markup that ideally have access to the containing components props.

ryanatkn:
I'll voice a niche usecase that I have only some experience with: using Svelte to "draw" with a bunch of HTML or SVG elements, like algorithmic art/graphics. For example see the Twitter hashtags #creativecoding and #generative: https://twitter.com/search?q=%23creativecoding%20OR%20%23generative%20&f=top (I hesitate to use this example because of recent _developments_ in the digital art world, but it's the best example I know of; ignore any related hashtags you dislike)

The idea here is that you're putting arbitrary stuff onscreen, feeling your way around with props/params that do mathy/algorithmic things, completely unlike normal business apps. In these situations it's common to have multiple reusable bits of ad-hoc logic/markup. Most stuff in the wild is probably generated with functions on a canvas, not a template language like Svelte, but Svelte works well for many of these things too, and extracting every reusable bit into a component is a burden.

eden-omb:
As a yet-additional use case, Storybook is currently quite annoying to work with in Svelte if you're using more than one template. I think about inline components as just private top-level components for internal use inside a given component. If there's cause for them to be scoped and not unique within a component (as the RFC asks) I believe it's only reasonable for those bits to be extracted to a different component.

As for context and styling, since these are for innercomponental use only I believe they should be inherited. This could likely be done with a preprocessor as @PierBover suggests. Reusable bits of markup inside a template would be massively helpful in simplifying templates and reducing duplication.

I will point out that this is very easy to do is something like JSX where you can simply keep bits of markup in variables and render them wherever you want, and being able to do that purely inside the template is an important core feature to have I believe. I think this should be built into Svelte as an internal processor.

arxpoetica:
I don't believe it's niche. I run into the need all the time. I think I'll have to remember to come comment on this, however, since—at the moment—I can't seem to pull them out of my head.

One thing that is vaguely coming to my memory is use cases where `{#if...}...{/if}` blocks get unnecessarily complex to include or exclude nested HTML that has to be repeated more than once. Usually, in those use cases, building a new component is annoying. This is sufficiently vague to not be a useful comment, so, like I said, I'll try and remember to come back to this.

Regardless, I want to push back on the thought that it's niche.

AlexxNB:
Just FYI. Maybe somebody knows that Malina is marginal framework similar to Svelte. It have [`fragments`](https://malinajs.github.io/docs/#fragment), a some sort of inline components which we discussing here.
When creating UI-lib for this framework I used fragments many times. They allow me to avoid code repeating  and leave template in current component. Separating template into child component is not logical here and it increase the bundle. 
https://github.com/AlexxNB/malina-ui/blob/master/components/Form/Button.xht
https://github.com/AlexxNB/malina-ui/blob/master/components/Form/Input.xht

I like the definition of the `fragments` by author of Malina's - '[It is a function, but for template](https://javascript.plainenglish.io/how-fragments-can-help-in-your-web-development-5efc4d10f9da)'.

PaulMaly:
In addition, in RactiveJS, it is called [Partials](https://ractive.js.org/tutorials/partials/). But I believe @Rich-Harris know about it )))

brunnerh:
Just opened [an issue](https://github.com/sveltejs/svelte/issues/7972) about a feature like this and was pointed here. For me a key point would be to get access to the component class for its use with the client-side component API. This aids in interoperability with non-svelte libraries/components.

My main suggestions summarized:
- Reuse `<svelte:fragment>` for the definitions
- `let:...` behaves like a property (`export let ...`)
- `bind:this={Component}` binds the component class to a local variable (in this case `Component`; has to be declared with `let`)
  - Allows use with client-side API (`new Component({ target: ... })`)
  - Allows use with `<svelte:component this={Component}>`
- `name="Component"` would allow direct use with a regular tag: `<Component />`

sken130:
Saying this feature is not needed is like saying Java should not have inner classes and C# should not have nested classes.

Even only 5% of people have encountered valid use cases for inline components, it's already enough to make it a valid feature.

Also, when persuading other developers and management to adopt a new framework like Svelte instead of React, sometimes, a single downside (apart from ecosystem being small) is enough for them to turn you down, despite having so many other advantages. And the lack of inline components is one of such downsides. This features may look small, but if you need it, then it is a critical feature. I want to use Svelte, but I don't know what argument I could make up for the lack for this feature.

KevsRepos:
I just want to mention, that this:

```svelte
<script>
	let importantMessage;
</script>

<template bind:this={importantMessage}>
	<h1>Cake</h1>
</template>

{@html importantMessage?.innerHTML}
```
is already possible. https://svelte.dev/repl/f8ebf7109afa4608a604f04bc13b640b?version=3.55.0
And I would leave it like that. *Basically*. We could implement syntactic sugar for this line: `{@html importantMessage?.innerHTML}` like `{@template importantMessage}`

I guess this would not allow for nested child elements... As of now, it is impossible to use the `slot` prop in `<template>` elements as Svelte hijacked the slot mechanism as we all know and errors out, if we try to use it in the original native way.

So what about changing `{@template importantMessage}` to `<svelte:template></svelte:template>` and allowing slots inside template tags to be used the way they were intended to use.

**Why use `<template>` instead of <svelte:*>?** 
"Use the platform" - and stick as near as possible to the platform. Svelte has always taken credit for feeling like "native html and javascript", though I couldnt agree less, not hijacking the meanings of the `<slot>` element and not bringing in *another* framework specific element or changing the meanings of already existing framework specifc elements, would actually make Svelte feel more like the native way of doing things.
Why bring in our own way of doing things, when html actually has a neat syntax already?

EDIT: Additionally I want to say that basically Svelte's strength is hiding the DOM API behind very semantic, declarative and neat syntax. So let us think the other way around: html and javascript already give us everything we need. Its only the DOM API which would require us the `template.cloneNode()` and so on. So its this specific part that we need to hide. And with `<svelte:template>`, we simply hide the implementation of cloneNode and so on.

I dont think we need scoped styles for templates. CSS classes are a thing! I think leaving the styles scoped to the document is enough. And eh voila, we kinda have reusable chunks of html.

Though I agree we might need a little bit more dynamic stuff here. So far, I could'nt find a solution to add the content of a template element with dynamic properties. It seems nearly impossible to do so currently in a reasonably usable way.

So what about adding a syntax to allow this? This is especially difficult to determine how syntax for that should look like. Any ideas?

space-nuko:
Is this going to be implemented? Ran across the need myself

pngwn:
@space-nuko the verdict is still out but not before the next major version. We'll need to set aside some time to go through the RFCs prior to planning the next version.

GermanJablo:
I've just added a detailed comment arguing for multiple components per file and offering a proof of concept of how I think the best syntax to achieve this would be here: https://github.com/sveltejs/svelte/issues/2940

I'd like to suggest taking a step back from this proposal to consider "components as functions". Basically the reasons why I consider this better than templates or any kind of wrapper on components are the following:

1. Using additional wrapping increases the number of characters and reduces the readability you've worked so hard to excel at. Using components as functions not only does not increase the number of characters, but in many cases reduces it (see my PoC above). In addition, using functions show additional valuable information in the components: namely, the name of the component.
2. I think it's easier to reason about component properties in terms of function parameters, rather than imports and exports. That's not to mention that in the world of JS, you can export constants or variables. An abstraction that does not translate well to functions.
3. This would make Svelte more similar to most frameworks. Migrating to Svelte from another framework would be easier, and adopting or learning it would look less intimidating. 
4. It is objectively simpler. This RFC would have to introduce more concepts and abstractions.
5. It is more powerful. From what I've read of this thread, much of the discussion centers around whether inline templates should include `<styles>` or `<script>` . In most frameworks, since the components are functions, it is possible to combine them in the same file without having the limitations that are being considered with this syntax.

dummdidumm:
This is solved in Svelte 5 through snippets, therefore closing this RFC.

PR #35, opened 2020-09-10
Remove Service Worker support from Sapper

benmccann:
[Rendered](https://github.com/benmccann/rfcs/blob/rm-service-workers/0000-remove-service-worker.md)

### DIFF
--- a/0000-remove-service-worker.md
+++ b/0000-remove-service-worker.md
@@ -1,0 +1,84 @@
+- Start Date: 2020-09-10
+- RFC PR: (leave this empty)
+- Svelte Issue: (leave this empty)
+
+# Remove service worker support from Sapper
+
+## Summary
+
+Remove the built-in service worker support from Sapper. Most users don't need
+service workers and should probably remove them from their applications. For the few
+that do need service workers, we can encourage them to include the service worker
+separately.
+
+## Motivation
+
+Service workers are primarily meant to support progressive web apps (PWAs), which
+are essentially replacements for native mobile applications. Service workers can be
+a valuable tool for building PWAs which work offline like a native mobile app would.
+However, the vast majority of sites built with Sapper do not fit this profile. Thus,
+including a service worker in the template is a very opinionated decision that is a
+poor fit for the majority of our users.
+
+The existing service worker support has shown several complications:
+
+* They very aggressively cache files
+    * Hard to diagnose issues can occur when deploying a new version of a site
+    * The caching continue to occur even with a hard refresh (i.e. Ctrl+F5).
+      Developers and users are not familiar with needing to unregister a service
+      worker in the Application tab in Chrome
+    * The caching happens across different sites during local development. If a
+      developer works on one site running at the default localhost:3000 and then
+      switches to working on another site that they also run on the default port
+      then strange issues will appear where content from the other site is served
+    * The application server can appear to be running when it is not
+    * If a user has a large amount of files in the `/static` directory (e.g. over
+      50mb) then the application can stop responding.
+* The default service worker downloads all `/client` files
+    * This can result in excessive bandwidth and usage charges for end users on
+      mobile devices
+    * This results in a worse score on Chrome's built-in Lighthouse performance
+      scoring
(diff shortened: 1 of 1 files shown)


arxpoetica:
Initially I was for this, but after some discussion w/ @pngwn I think I'm more inclined to just fix all the vexing problems around our current SWs setup, rather than tear it out all together.

The main thing that needs firm shoring up is controlled cache invalidation. Uncertain if timestamps are sufficient for this. Might need to be programmatic w/ a nonce flag. Dunno. SWs are certainly weird, which I think is basically (more) what this RFC is about. Fixing the weirdness.

TheComputerM:
I think this should be fine as long as we include a tutorial (preferably in the sapper FAQ) on how to add a service worker.

jbg:
@arxpoetica Fixing the SWs would be a great idea, but I strongly suggest removing them from the standard Sapper templates. Most people building a Sapper app don't need a SW, and they're a regular source of hard-to-debug problems. The very first thing I do when starting a Sapper app is rip out the SW.

arxpoetica:
Regarding this statement in the RFP:

> None or almost none of the core Sapper developers use service workers themselves.

Guilty. But I don't think that's a good reason to rip them out. Rather, I favor:

* Minimizing the default cache -> Have the SW present, but opt in on what to cache, and document how to opt in. Flip aggressive caching on its head, and instead aggressively don't cache.
* Provide an easier mechanism/API for doing certain tasks that normally require looking up the SW API. For example, cache clearing/busting. Let's load up some tooling/API around that.
* Document better and perhaps as part of the installation that it's THERE and if people don't want it they can remove it. That is, don't bury the lede, so to speak.
* This is API again, and more just something to be brainstormed, but PWA really are capable of a ton of "offline" things, and we probably should discuss more the pros & cons of putting this kind of API more front and center in terms of the promise of "Sapper" as a PWA. My inclination is to think that people don't realize what's possible, don't have the tooling, and so, generally, don't do it.

lukeed:
I use SWs – just not Sapper's. Purposefully keep my SW usage relatively simple.

benmccann:
@lukeed can you share an example of your service worker? I'd like to see what a better implementation would look like

lukeed:
I plan to release an OSS app that would include a template for this. Currently I can't since they're tied up in private projects & not all that easy to extract at the moment... but 🔜 

I haven't checked in on this thread in a while – do we have a definitive list as to _why/what_ needs fixing?

benmccann:
I tried to include a list of issues in the RFC: https://github.com/benmccann/rfcs/blob/rm-service-workers/0000-remove-service-worker.md#motivation

pngwn:
If we remove the eager caching of all assets and do it selectively in an install step, half the issues with sapper's sw will go away. The DX with service workers isn't ideal but we can add a note in the docs about that. 

Sapper's service worker implementation isn't ideal, I personally do use service workers (heavily in some cases) and am familiar with the APIs and common approaches. However, I don't really use Sapper. If I used it more, I would modify the SW substantially in those cases.

thgh:
I would prefer to keep support, but disable the service worker by default. Vue nails this with the [PWA plugin](https://cli.vuejs.org/core-plugins/pwa.html).

seanlail:
Completely agree with @thgh and others here - disabled by default would be great.

Both Vue and React (at least CRA) have an opt-in strategy.
https://cli.vuejs.org/core-plugins/pwa.html
https://create-react-app.dev/docs/making-a-progressive-web-app/#why-opt-in

I quite like the Vue consideration around development too (see the noopServiceWorker).
> Instead, in the development mode the noopServiceWorker.js is included. This service worker file is effectively a 'no-op' that will reset any previous service worker registered for the same host:port combination.

As mentioned, there are current issues around service workers in Sapper, example:
https://github.com/sveltejs/sapper/issues/1661

So opt-in would be welcomed.

Perhaps a short-term "fix" is to document how to disable the current service worker implementation?

benmccann:
Closing this since we've moved on from Sapper to SvelteKit

PR #36, opened 2020-09-16
SvelteKit's router as a standalone component

benmccann:
[Rendered](https://github.com/benmccann/rfcs/blob/routing/text/0000-routing.md)


madeleineostoja:
I think this is a really important RFC for the wider community. Svelte _needs_ a blessed routing solution that isn't bolted onto Sveltekit, especially for SPAs where Sveltekit is overkill / cumbersome.

Personally I would have loved if that router was declarative rather than file-based (I love what [tinro](https://github.com/AlexxNB/tinro) is doing), but taking my own preferences out of it having a routing solution semi-officially supported by Svelte is a must-have and an oversight in the current ecosystem.

nikkwong:
This is desperately desperately needed. I have a standard application where there are two types of users `students` and `teachers`. Right now, because of file based routing; I have to push each of their homepages to `/student` and `/teacher` rather than at the base route `/`. 

Using `src/routes/index.svelte` and then in that component using `<svelte:component this="{isTeacher ? TeacherHome : StudentHome})">` as a workaround is not a good solution because then I can't take advantage of the `load` function for each component. Currently there is no workaround; so aliases are desperately needed. Very much looking forward to this PR.

CanRau:
might also solve https://github.com/sveltejs/kit/pull/1130

axljoker03 (approved):

benmccann:
@axljoker03 please stop approving all the RFCs. You're spamming all the maintainers and it's a bit annoying. You can simply leave a thumbs up or down on the post to indicate an opinion

geekox86:
Greetings all. May I ask about the status of this RFC, as my team is following a specific project directory structure that suits our needs and conventions but it is highly incompatible with the file-based routing. It would be super to be able to define the routes and sub-routes at each nested level of the UI layout using other approaches.

My objective is to be able to use the SSR offered by SvelteKit without the file-based router. Our backend API is not NodeJs and we would like to communicate directly with it from the frontend.

designermonkey:
I would like to see routing separated from the core kit as, unfortunately, it prevents us from using the Svelte kit. File-based routing goes against everything we do with modularised code, where each high-level component (root level route) is fully isolated and defines its routes that are added to an app router as modules. We can't switch our methodology to file-based routing without massive pain, as we would have to change how we work.

If routing were separate, with a well-defined API so alternative solutions could be produced that would be compatible with the svelte kit, we would start to use it.

PR #37, opened 2020-09-18
Strongly Typed Svelte Component Interface

dummdidumm:
[Rendered](https://github.com/dummdidumm/rfcs/blob/svelte-component-interface/text/svelte-component-interface.md)

### DIFF
--- a/text/svelte-component-interface.md
+++ b/text/svelte-component-interface.md
@@ -82,7 +82,7 @@ The Svelte component definition is enhanced with typings for Props, Events, Slot
 
 Obviously it's not ideal to have type-only-properties on the definition, but with a big enough disclaimer/comment on it, this should be okay in my opinion, especially since the `$$`-prefix is already the "internal, don't use this"-notion for other properties/methods.
 
-Regarding "where to put this": This is up to discussion. Currently I think enhancing `SvelteComponentDef` (which is `SvelteComponent` on the outside) with that definition is the best option. This would not be a breaking change to the existing definition because of the `any` type by default for props/events/slots.
+Regarding "where to put this": This is up to discussion. Currently I think enhancing `SvelteComponentDev` (which is `SvelteComponent` on the outside) with that definition is the best option. This would not be a breaking change to the existing definition because of the `any` type by default for props/events/slots.
 
 ## How we teach this
 
--- a/text/svelte-component-interface.md
+++ b/text/svelte-component-interface.md
@@ -30,7 +30,7 @@ import { SvelteComponent } from 'svelte';
 declare module "myModule" {
   export class MyComponent extends SvelteComponent<
     { propA: boolean },
-    { someEvent: string },
+    { someEvent: CustomEvent<string>; click: MouseEvent },
     { default: { aSlot: number } }
   > {}
 }
@@ -71,7 +71,7 @@ class Svelte2TsxComponent<
 
   $on<K extends keyof Events>(
     event: K,
-    handler: (e: Events[K]) => any
+    handler: (e: Events[K]) => void
   ): () => void; // <-- typed
   $set(props: Partial<Props> & Record<string, any>): void; // <-- typed, Record<string, any> for the $$restProps possibility
   // ...
--- a/text/svelte-component-interface.md
+++ b/text/svelte-component-interface.md
@@ -6,7 +6,7 @@
 
 ## Summary
 
-Provide a Svelte component type definition which library authors can use to provide a strongly typed Svelte component which the language server understands:
+Provide a Svelte component type definition which library authors can use to provide a strongly typed Svelte component which the language server understands. Inside a `.d.ts` file:
 
 ```ts
 import { SvelteComponent } from "svelte";
@@ -22,19 +22,27 @@ Currently there is no official and easy way to provide type definitions. Devs wh
 
 ### Usage would look like this:
 
+Inside a `d.ts` file:
+
 ```ts
 import { SvelteComponent } from 'svelte';
 
-export class MyComponent extends SvelteComponent<
-  { propA: boolean },
-  { someEvent: string },
-  { default: { aSlot: number } }
-> {}
+declare module "myModule" {
+  export class MyComponent extends SvelteComponent<
+    { propA: boolean },
+    { someEvent: string },
+    { default: { aSlot: number } }
+  > {}
+}
 ```
 
 This type definition would result in this being correct usage of the component:
 
 ```
+<script>
+  import { MyComponent } from "myModule";
+</script>
+
 <MyComponent propA={true} on:someEvent={e => e.detail.toUpperCase()} let:aSlot={aNumber} />
 ```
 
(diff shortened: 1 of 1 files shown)


benmccann:
One thing I noticed in Sapper is that the codebase never actually got type-checked despite using TypeScript because it was using the sucrase plugin, which skips type checking. A lot of the types were wrong as a result

It looks like svelte mostly uses sucrase as well, though uses the typescript plugin when publishing. If we're going to start publishing more types, a good first step would be to use the typescript plugin at least when running the tests on the CI server so that we can make sure when we check in code that we're conforming to the types we're writing. In Sapper I used `rollup-plugin-typescript2` because it's twice as fast and you can toggle type-checking on and off depending on which npm task you're running (vs toggling between `@rollup/plugin-sucrase` and `@rollup/plugin-typescript` based on whether you want typechecking)

Also, if we're going to type the components, one thing to mention is that I think we would need separate type definitions for the DOM and SSR Components.

dummdidumm:
> It looks like svelte mostly uses sucrase as well, though uses the typescript plugin when publishing. If we're going to start publishing more types, a good first step would be to use the typescript plugin at least when running the tests on the CI server so that we can make sure when we check in code that we're conforming to the types we're writing.

I looked at the tests and they are all using JS only, so no point in using TS for type checking there. Since the build does type checking, I think we're save.

> Also, if we're going to type the components, one thing to mention is that I think we would need separate type definitions for the DOM and SSR Components.

This RFC is only about enhancing the typing in such a way that you can use to author types. Adding the type definition for SSR could be just adding a static "render" method to the type definition.

benmccann:
> Since the build does type checking, I think we're save.

I think the build only does type-checking when cutting a release though? It'd be nice to do it more often to catch errors before then

benmccann:
I converted a lot of Svelte's tests to TypeScript here: https://github.com/sveltejs/svelte/pull/5433

nateshmbhat:
any ETA on when this can be merged ?

orta:
Re: 

> The `$$prop_def`/`$$events_def`/`$$slot_def` leak.

I'm not sure how I see those leaking? They look completely internal to the language server to me from this RFC.

My assumptions: As ever, so far I've managed to make a svelte app with 2 components, so consider my take naive. However, I did read the code for [carbon-components-svelte](https://unpkg.com/browse/carbon-components-svelte@0.16.0/) who is already using the `$$` variables ( sneaky but cool @metonym )

I assume that every .svelte file would have an equivalent `.d.ts`: e.g.

```html
// Answer.svelte
<script>
	export let answer;
</script>

<p>The answer is {answer}</p>
```

Would need something like:

```ts
// Answer.d.ts
import { SvelteComponent } from 'svelte';
export default class Answer extends SvelteComponent<{ answer: string }, { }, { }> {}
```

This would also get scooped up probably into a root module export which would look something like:

```ts
import { SvelteComponent } from 'svelte';

module "testing" {
   export class Answer extends SvelteComponent<{ answer: string }, { }, { }> {}
   export class Question extends SvelteComponent<{ question: string, subtitle: string }, { }, { }> {}
}
```

Where I think this RFC could still do with a bit of work: I think this could be mostly tool powered too, could a rollup plugin exist for generating the svelte types by asking for info (in this format) from the language service?

dummdidumm:
> Re:
> 
> > The `$$prop_def`/`$$events_def`/`$$slot_def` leak.
> 
> I'm not sure how I see those leaking? They look completely internal to the language server to me from this RFC. Well, I guess you could look inside the svelte .d.ts to see them.

True. My worries were/are that those are purely for type checking and don't exist at runtime. But they would turn up in the autocompletion when someone uses the the instance of that class type in a `script` tag. But we could in language-tools filter out these completions. At that point it's like you said, someone would have to look this up to actually find it. 

> Where I think this RFC could still do with a bit of work: I think this could be mostly tool powered too, could a rollup plugin exist for generating the svelte types by asking for info (in this format) from the language service?

This is a second step after this one is agreed on (which looks like it almost is). As you said, I envision some kind of small package which could do this, which then could get wrapped by a rollup plugin.

orta:
Alright, looks like we've got general consensus that this is the way forwards 👍🏻 - I'll merge tomorrow unless we get any other feedback

nateshmbhat:
Waiting patiently to be merged 🧘‍♂️

firefish5000:
```ts
Props extends {} = any
```
Any reason not to use 
```ts
Props extends Record<string,any> = Record<string,any>
```
 instead? Just to ensure type errors are caught while typeing the defs so if someone types something like
```ts
Svelte2TsxComponent<'name' ,2 ,(evt: any)=>void>
``` 
it will be caught as an error right away rather than waiting for someone (possibly downstream) to try to use the component and report the error.

dummdidumm:
In the implementing PR I actually did `Props extends Record<string, any> = any`, I'll update the RFC to match it.

orta:
Agree, good feedback 👍🏻 OK - I"m going to hit the merge button. Thanks all!

TheComputerM:
Where should be the documentation on how to use this?

nateshmbhat:
By when can this be published on npm?

vatro:
Hi!
I just spent quite a while to figure something out and I think this is a good place to share, because it involves `SvelteComponentTyped` and `$$prop_def`. Both were essential for the solution I came up with and I'd definitely like to keep `$$prop_def` and other `$$` around and accessible ( especially if there're no alternatives (?)🤔 )

Consider following scenario:
Trying to add additional props to a Svelte-TypeScript component's interface without declaring them inside the `<script>`-tag. These additional props can either be "manually" defined or derived / modified from some other class / type (_and finally processed by the component via `$$props` or `$$restProps` references_):

```svelte
<!-- Test.svelte -->
<script lang="ts">

  export let fooScript: number = undefined
  export let barScript: number = undefined
  export let bazScript: number = undefined
  
  export function testPublicFunction(): void {
      console.log("testPublicFunction!")
  }
</script>
```

```typescript
// Test.d.ts
import Test_svelte_ts from "./Test.svelte"

type additionalProps = {
    foo?: number
    bar?: string
    baz?: boolean
}

export default class Test extends SvelteComponentTyped< typeof Test_svelte_ts.prototype.$$prop_def & additionalProps, {}, {}> {}
```
_Hover over component in VS Code:_
image

☝️ adding ...
```typescript
typeof Test_svelte_ts.prototype.$$prop_def
```
did the trick for me accessing (exported) props declared inside the `<script>`-tag (`Test.svelte` file).
Without it ...
```typescript
export default class Test extends SvelteComponentTyped<additionalProps, {}, {}> {}
```
_Hover over component in VS Code:_
image
Only `additionalProps` get listed on hover + type errors if you try to add one of the props declared in the `Test.svelte` file.

❓🤔 one thing I then stumbled upon is that `$$events_def` & `$$slot_def` are not defined in `SvelteComponentDev`, so I don't know yet how to inject Events and Slots of the base component into those other two parameters of the `SvelteComponentTyped` `<{ ok }, { ? }, { ? }>` class:

see `node_modules\svelte\types\runtime\internal\dev.d.ts` ( Svelte 3.31.2  ) 
image

Please correct me if I'm wrong (somewhere) or if there's a better approach to achieve the above. Thanks! 👍

PR #38, opened 2020-09-20
TypeScript: Explicitly Typing Props/Slots/Events + Generics

dummdidumm:
[rendered](https://github.com/dummdidumm/rfcs/blob/ts-typedefs-within-svelte-components/text/ts-typing-props-slots-events.md)

### DIFF
--- a/text/ts-typing-props-slots-events.md
+++ b/text/ts-typing-props-slots-events.md
@@ -6,7 +6,7 @@
 
 ## Summary
 
-Provide possibilites for TypeScript users to strongly type a Svelte component's props/events/slots, including generics. For that, we introduce reserved interfaces named `ComponentProps`, `ComponentEvents`, `ComponentSlots`, `ComponentDef` to type parts or all of them at once. We also introduce `<script>` attributes for generics and for marking a component as having no other events besides the ones defined within.
+Provide possibilites for TypeScript users to strongly type a Svelte component's props/events/slots, including generics. For that, we introduce reserved interfaces named `ComponentProps`, `ComponentEvents`, `ComponentSlots`. We also introduce a concept for generics and a `<script>` attribute for marking a component as having no other events besides the ones defined within.
 
 While this is not a change to Svelte's core, it's still something that needs to be specified so intellisense implementers have something to adhere to.
 
@@ -17,6 +17,7 @@ Using TypeScript with Svelte provides a lot of goodness already, but there are s
 - There is no way currently to tell the intellisense that there's only a fixed set of events one can listen to. You can type `createEventDispatcher` but that does still make it possible to listen to other events
 - There is no way currently to explicitely type slots
 - There is no way currently to use generics
+- There is no way currently to make a component implement some specified interface and have this checked by types
 
 ## Detailed design
 
@@ -118,7 +119,24 @@ This works the same as for typing events. You probably won't use that because it
 
 You want to specify some generic connection between props/slots/events. For example you have a component which has an input prop `item`, and an event called `itemChanged`. You want to use this component for arbitrary kinds of item, but you want to make sure that the types for `item` and `itemChanged` are the same. Generics come in handy then. You can read more about them on the [official TypeScript page](https://www.typescriptlang.org/docs/handbook/generics.html).
 
-##### Option 1
+#### Solution
+You use new reserved type called `ComponentGeneric`.
+
+```html
+<script lang="ts">
+    import {createEventDispatcher} from "svelte";
+
+    type T = ComponentGeneric<boolean>; // extends boolean
+    type X = ComponentGeneric; // any
+
+    export let array1: T[];
+    export let item1: T;
+    export let array2: X[];
+    const dispatch = createEventDispatcher<{arrayItemClick: X}>();
+</script>
+```
+
+##### Discarded alternative 1
 You use a new `<script>` attribute called `generics`. The contents of that attribute have to be valid generic typings.
 
 ```html
@@ -135,7 +153,9 @@ You use a new `<script>` attribute called `generics`. The contents of that attri
 ...
 ```
 
-##### Option 2
+Discarded because it is invalid TypeScript without additional transformations.
+
+##### Discarded alternative 2
 You use a new reserved interface called `ComponentGenerics` and do the typing on it, not declaring any properties on it.
 
 ```html
@@ -154,24 +174,10 @@ You use a new reserved interface called `ComponentGenerics` and do the typing on
 ...
 ```
 
-#### Option 3
-You use new reserved type called `ComponentGeneric`.
+Discarded because it is invalid TypeScript without additional transformations.
 
-```html
-<script lang="ts">
-    import {createEventDispatcher} from "svelte";
 
-    type T = ComponentGeneric<boolean>; // extends boolean
-    type X = ComponentGeneric; // any
-
-    export let array1: T[];
-    export let item1: T;
-    export let array2: X[];
-    const dispatch = createEventDispatcher<{arrayItemClick: X}>();
-</script>
-```
-
-### ComponentDef
+### ComponentDef (likely discarded)
 
 If you want to type all at once, because you like to have the definition in one place or want to better define a generic relationship, you can use the `ComponentDef` interface.
 
@@ -188,7 +194,9 @@ If you want to type all at once, because you like to have the definition in one 
 ...
 ```
 
-### ComponentDef alternative: Namespace
+This is likely not a good idea because you then can achieve typing a component in multiple ways, which introduces maintenance overhead for implementations.
+
+#### Discarded ComponentDef alternative: Namespace
 
 As an alternative to the `ComponentDef` interface, one could use a namespace and put the interfaces inside it. That would make refactoring easier if you for example start of with typing only the events but want to add more typings to slots later on. This would come at the cost of uncanny-valley-stuff for defining the generics.
 
@@ -216,6 +224,8 @@ As an alternative to the `ComponentDef` interface, one could use a namespace and
 ...
 ```
 
+This is discarded because it provides no real benefit over the three-seperate-interfaces-solution.
+
 ### Summary
 
 As you can see, there would be several options to achieve the same. You can use `ComponentDef` to type all at once, or you can mix and match the other possibilities to only type part of it. The drawback is that there is more than one way to achieve the same goal. But only having `ComponentDef` may be too much typing overhead of you only want to specifically type parts of the component. In general, props, slots and their types are already inferable quite nicely at this point. Only generics and events are where you really would need this.
@@ -233,12 +243,13 @@ For users: Enhance docs. For intellisense devs: A more formal specification outl
 - This will only work for TS users
 - Uncanny-valley-stuff for generics
 - Reserved interface names could collide with existing ones, but I think that's rare. It's also only a breaking change for the language-tools because it does not affect the core of Svelte
-- Many ways to achieve the same goal (interface/generics combinations)
+- If we implement `SvelteComponentDef`, then there are multiple ways to achieve the same goal (interface/generics combinations)
 
 ## Alternatives
 
 - Don't do anything and say "well, there are some limits". VueJS for example also cannot deal with generics as far as I know.
-- Only provide parts of this solution: `strictEvents` and `generics`, but from the interfaces only `ComponentDef`, and tell people "if you want to type it, type it all".
+- Only provide parts of this solution: `strictEvents` and generics, and from the interfaces only `ComponentDef`, and tell people "if you want to type it, type it all".
+- Shorter interface name alternatives: `Props` / `Events` / `Slots` / `Generic`. More likely that they clash with existing definitions?
 
 ## Unresolved questions
 
--- a/text/ts-typing-props-slots-events.md
+++ b/text/ts-typing-props-slots-events.md
@@ -116,11 +116,33 @@ This works the same as for typing events. You probably won't use that because it
 
 ### Generics
 
-You want to specify some generic connection between props/slots/events. For this you use the `<script>` attribute `generics`. The contents of that attribute have to be valid generic typings.
+You want to specify some generic connection between props/slots/events. For example you have a component which has an input prop `item`, and an event called `itemChanged`. You want to use this component for arbitrary kinds of item, but you want to make sure that the types for `item` and `itemChanged` are the same. Generics come in handy then. You can read more about them on the [official TypeScript page](https://www.typescriptlang.org/docs/handbook/generics.html).
+
+##### Option 1
+You use a new `<script>` attribute called `generics`. The contents of that attribute have to be valid generic typings.
 
 ```html
 <script lang="ts" generics="T extends boolean, X">
     import {createEventDispatcher} from "svelte";
+
+    export let array1: T[];
+    export let item1: T;
+    export let array2: X[];
+
+    const dispatch = createEventDispatcher<{arrayItemClick: X}>();
+</script>
+
+...
+```
+
+##### Option 2
+You use a new reserved interface called `ComponentGenerics` and do the typing on it, not declaring any properties on it.
+
+```html
+<script lang="ts">
+    import {createEventDispatcher} from "svelte";
+    
+    interface ComponentGenerics<T extends boolean, X> {}
 
     export let array1: T[];
     export let item1: T;
@@ -143,6 +165,34 @@ If you want to type all at once, because you like to have the definition in one 
         props: { items: T[]; someOptionalProp?: string; };
         events: { itemClick: CustomEvent<T>; };
         slots: { default: { item: T; }; };
+    }
+</script>
+
+...
+```
+
+### ComponentDef alternative: Namespace
+
+As an alternative to the `ComponentDef` interface, one could use a namespace and put the interfaces inside it. That would make refactoring easier if you for example start of with typing only the events but want to add more typings to slots later on. This would come at the cost of uncanny-valley-stuff for defining the generics.
+
+```html
+<script lang="ts">
+    // ...
+    declare namespace Component {
+      interface Generics<T> {}
+    
+      interface Events {
+        itemClick: CustomEvent<T>; // use generic T here which will be the one defined in Generics
+      }
+    
+      interface Props {
+        itemClick: CustomEvent<T>;
+        someOptionalProp?: string;
+      }
+    
+      interface Slots {
+        default: { item: T }
+      }
     }
 </script>
 
--- a/text/ts-typing-props-slots-events.md
+++ b/text/ts-typing-props-slots-events.md
@@ -154,6 +154,23 @@ You use a new reserved interface called `ComponentGenerics` and do the typing on
 ...
 ```
 
+#### Option 3
+You use new reserved type called `ComponentGeneric`.
+
+```html
+<script lang="ts">
+    import {createEventDispatcher} from "svelte";
+
+    type T = ComponentGeneric<boolean>; // extends boolean
+    type X = ComponentGeneric; // any
+
+    export let array1: T[];
+    export let item1: T;
+    export let array2: X[];
+    const dispatch = createEventDispatcher<{arrayItemClick: X}>();
+</script>
+```
+
 ### ComponentDef
 
 If you want to type all at once, because you like to have the definition in one place or want to better define a generic relationship, you can use the `ComponentDef` interface.
--- a/text/ts-typing-props-slots-events.md
+++ b/text/ts-typing-props-slots-events.md
@@ -1,0 +1,181 @@
+- Start Date: 2020-09-20
+- RFC PR: 
+- Svelte Issue: https://github.com/sveltejs/language-tools/issues/442
+
+# TypeScript: Typing Props/Events/Slots + Generics
+
+## Summary
+
+Provide possibilites for TypeScript users to strongly type a Svelte component's props/events/slots, including generics. For that, we introduce reserved interfaces named `ComponentProps`, `ComponentEvents`, `ComponentSlots`, `ComponentDef` to type parts or all of them at once. We also introduce `<script>` attributes for generics and for marking a component as having no other events besides the ones defined within.
+
+While this is not a change to Svelte's core, it's still something that needs to be specified so intellisense implementers have something to adhere to.
+
+## Motivation
+
+Using TypeScript with Svelte provides a lot of goodness already, but there are some areas which lack support:
+
+- There is no way currently to tell the intellisense that there's only a fixed set of events one can listen to. You can type `createEventDispatcher` but that does still make it possible to listen to other events
+- There is no way currently to explicitely type slots
+- There is no way currently to use generics
+
+## Detailed design
+
+### Typing events
+
+Use case: You want to strictly type events. Listening to anything outside the defined events should throw a type error.
+
+You start with one event which is from your own typed `createEventDispatcher` and one forwarded event.
+
+```html
+<script lang="ts">
+    import {createEventDispatcher} from "svelte";
+
+    const dispatch = createEventDispatcher<{own: boolean}>();
+</script>
+
+<button on:click={() => dispatch('own', true)}>Own</button>
+<button on:click>Forwarded</button>
+```
+
+Now you want to ensure that listeing to anything else than `on:own`/`on:click` throws a type error. For that you use the new `<script>` attribute `strictEvents`:
+
+```html
+<script lang="ts" strictEvents>
+    import {createEventDispatcher} from "svelte";
+
+    const dispatch = createEventDispatcher<{own: boolean}>();
+</script>
+
+<button on:click={() => dispatch('own', true)}>Own</button>
+<button on:click>Forwarded</button>
+```
+
+Now you add one event which comes from a dispatcher mixin:
+
+```html
+<script lang="ts" strictEvents>
+    import {mixinDispatch} from "somewhere";
+    import {createEventDispatcher} from "svelte";
+
+    const dispatch = createEventDispatcher<{own: boolean}>();
+</script>
+
+<button on:click={() => mixinDispatch.mixinEvent('foo')}>Mixin</button>
+<button on:click={() => dispatch('own', true)}>Own</button>
+<button on:click>Forwarded</button>
+```
+
+In this case `strictEvents` will not work anymore because we cannot know that `mixinDispatch` dispatches events. So now you use the `ComponentEvents` interface.
+
+```html
+<script lang="ts">
+    import {mixinDispatch} from "somewhere";
+    import {createEventDispatcher} from "svelte";
+
+    interface ComponentEvents {
+        mixinEvent: CustomEvent<string>;
+        own: CustomEvent<boolean>;
+        click: MouseEvent;
+    }
+
+    const dispatch = createEventDispatcher<{own: boolean}>();
+</script>
+
+<button on:click={() => mixinDispatch.mixinEvent('foo')}>Mixin</button>
+<button on:click={() => dispatch('own', true)}>Own</button>
+<button on:click>Forwarded</button>
+```
+
+### Typing Slots
+
+This works the same as for typing events.
+
+```html
+<script lang="ts">
+    interface ComponentSlots {
+        default: { prop: boolean; };
+    }
+</script>
+
+<slot prop={true}></slot>
+```
+
+### Typing Props
+
+This works the same as for typing events. You probably won't use that because it's essentially doing the type work twice.
+
+```html
+<script lang="ts">
+    interface ComponentProps {
+        prop: boolean;
+    }
+
+    export let prop: boolean;
+</script>
+```
+
+### Generics
+
+You want to specify some generic connection between props/slots/events. For this you use the `<script>` attribute `generics`. The contents of that attribute have to be valid generic typings.
+
+```html
+<script lang="ts" generics="T extends boolean, X">
+    import {createEventDispatcher} from "svelte";
+
+    export let array1: T[];
+    export let item1: T;
+    export let array2: X[];
+
+    const dispatch = createEventDispatcher<{arrayItemClick: X}>();
+</script>
+
+...
+```
+
+### ComponentDef
+
+If you want to type all at once, because you like to have the definition on one place or want to better define a generic relationship, you can use the `ComponentDef` interface.
+
+```html
+<script lang="ts">
+    // ...
+    interface ComponentDef<T> {
+        props: { items: T[]; someOptionalProp?: string; };
+        events: { itemClick: CustomEvent<T>; };
+        slots: { default: { item: T; }; };
+    }
+</script>
+
+...
+```
+
+You can also leave out props/events/slots in that interface depending on what you actually use in your component.
+
+### Summary
+
+As you can see, there would be several options to achieve the same. You can use `ComponentDef` to type all at once, or you can mix and match the other possibilities to only type part of it. The drawback is that there is more than one way to achieve the same goal. But only having `ComponentDef` would be too much typing overhead of you only want to specifically type parts of the component. In general, props, slots and their types are already inferable quite nicely at this point. Only generics and events are where you really would need this.
+
+### Implementation hurdles
+
+We would need to make sure that we can provide some meaningful errors if the definition and the actual types don't match. So if someone types `ComponentSlots` as `{foo: boolean;}` but does `<slot foo={'aString'}></slot>`, we must highlight that. I have not looked closely into how this can be achieved yet because I want to first have agreement on the API.
+
+## How we teach this
+
+For users: Enhance docs. For intellisense devs: A more formal specification outlining the details.
+
+## Drawbacks
+
+- This will only work for TS users
+- Uncanny-valley-stuff for generics
+- Reserved interface names could collide with existing ones, but I think that's rare. It's also only a breaking change for the language-tools because it does not affect the core of Svelte
+- Many ways to achieve the same goal (interface/generics combinations)
+
+## Alternatives
+
+- Don't do anything and say "well, there are some limits". VueJS for example also cannot deal with generics as far as I know.
+- Only provide parts of this solution: `strictEvents` and `generics`, but from the interfaces only `ComponentDef`, and tell people "if you want to type it, type it all".
+
+## Unresolved questions
+
+- Interface wording ok?
+- Attribute wording ok?
--- a/text/ts-typing-props-slots-events.md
+++ b/text/ts-typing-props-slots-events.md
@@ -128,6 +128,11 @@ You use new reserved type called `ComponentGeneric`.
 
     type T = ComponentGeneric<boolean>; // extends boolean
     type X = ComponentGeneric; // any
+    
+    // you can use generics inside the other interfaces
+    interface ComponentSlots {
+        default: { aSlot: T }
+    }
 
     export let array1: T[];
     export let item1: T;
--- a/text/ts-typing-props-slots-events.md
+++ b/text/ts-typing-props-slots-events.md
@@ -115,6 +115,8 @@ This works the same as for typing events. You probably won't use that because it
 </script>
 ```
 
+If you define `$$Props`, all possible props need to be part of it. If you use `$$props` or `$$restProps` then that does _not_ widen the type, still only those defined in `$$Props` are allowed.
+
 ### Generics
 
 You want to specify some generic connection between props/slots/events. For example you have a component which has an input prop `item`, and an event called `itemChanged`. You want to use this component for arbitrary kinds of item, but you want to make sure that the types for `item` and `itemChanged` are the same. Generics come in handy then. You can read more about them on the [official TypeScript page](https://www.typescriptlang.org/docs/handbook/generics.html).
--- a/text/ts-typing-props-slots-events.md
+++ b/text/ts-typing-props-slots-events.md
@@ -122,6 +122,24 @@ If you define `$$Props`, all possible props need to be part of it. If you use `$
 You want to specify some generic connection between props/slots/events. For example you have a component which has an input prop `item`, and an event called `itemChanged`. You want to use this component for arbitrary kinds of item, but you want to make sure that the types for `item` and `itemChanged` are the same. Generics come in handy then. You can read more about them on the [official TypeScript page](https://www.typescriptlang.org/docs/handbook/generics.html).
 
 #### Solution
+
+You use a new `<script>` attribute called `generics`. The contents of that attribute have to be valid generic typings.
+
+```html
+<script lang="ts" generics="T extends boolean, X">
+    import {createEventDispatcher} from "svelte";
+
+    export let array1: T[];
+    export let item1: T;
+    export let array2: X[];
+
+    const dispatch = createEventDispatcher<{arrayItemClick: X}>();
+</script>
+
+...
+```
+
+##### Discarded alternative 1
 You use new reserved type called `$$Generic`.
 
 ```html
@@ -143,24 +161,7 @@ You use new reserved type called `$$Generic`.
 </script>
 ```
 
-##### Discarded alternative 1
-You use a new `<script>` attribute called `generics`. The contents of that attribute have to be valid generic typings.
-
-```html
-<script lang="ts" generics="T extends boolean, X">
-    import {createEventDispatcher} from "svelte";
-
-    export let array1: T[];
-    export let item1: T;
-    export let array2: X[];
-
-    const dispatch = createEventDispatcher<{arrayItemClick: X}>();
-</script>
-
-...
-```
-
-Discarded because it is invalid TypeScript without additional transformations.
+Discarded because it is invalid TypeScript when using advanced typings such as `const T extends string` and it's harder to read. The `generics` attribute reads exactly like the generics on a function.
 
 ##### Discarded alternative 2
 You use a new reserved interface called `$$Generics` and do the typing on it, not declaring any properties on it.
--- a/text/ts-typing-props-slots-events.md
+++ b/text/ts-typing-props-slots-events.md
@@ -37,7 +37,7 @@ You start with one event which is from your own typed `createEventDispatcher` an
 <button on:click>Forwarded</button>
 ```
 
-Now you want to ensure that listeing to anything else than `on:own`/`on:click` throws a type error. For that you use the new `<script>` attribute `strictEvents`:
+Now you want to ensure that listening to anything else than `on:own`/`on:click` throws a type error. For that you use the new `<script>` attribute `strictEvents`:
 
 ```html
 <script lang="ts" strictEvents>
@@ -134,7 +134,7 @@ You want to specify some generic connection between props/slots/events. For this
 
 ### ComponentDef
 
-If you want to type all at once, because you like to have the definition on one place or want to better define a generic relationship, you can use the `ComponentDef` interface.
+If you want to type all at once, because you like to have the definition in one place or want to better define a generic relationship, you can use the `ComponentDef` interface.
 
 ```html
 <script lang="ts">
@@ -149,11 +149,9 @@ If you want to type all at once, because you like to have the definition on one 
 ...
 ```
 
-You can also leave out props/events/slots in that interface depending on what you actually use in your component.
-
 ### Summary
 
-As you can see, there would be several options to achieve the same. You can use `ComponentDef` to type all at once, or you can mix and match the other possibilities to only type part of it. The drawback is that there is more than one way to achieve the same goal. But only having `ComponentDef` would be too much typing overhead of you only want to specifically type parts of the component. In general, props, slots and their types are already inferable quite nicely at this point. Only generics and events are where you really would need this.
+As you can see, there would be several options to achieve the same. You can use `ComponentDef` to type all at once, or you can mix and match the other possibilities to only type part of it. The drawback is that there is more than one way to achieve the same goal. But only having `ComponentDef` may be too much typing overhead of you only want to specifically type parts of the component. In general, props, slots and their types are already inferable quite nicely at this point. Only generics and events are where you really would need this.
 
 ### Implementation hurdles
 
--- a/text/ts-typing-props-slots-events.md
+++ b/text/ts-typing-props-slots-events.md
@@ -6,7 +6,7 @@
 
 ## Summary
 
-Provide possibilites for TypeScript users to strongly type a Svelte component's props/events/slots, including generics. For that, we introduce reserved interfaces named `ComponentProps`, `ComponentEvents`, `ComponentSlots`. We also introduce a concept for generics and a `<script>` attribute for marking a component as having no other events besides the ones defined within.
+Provide possibilites for TypeScript users to strongly type a Svelte component's props/events/slots, including generics. For that, we introduce reserved interfaces named `$$Props`, `$$Events`, `$$Slots`. We also introduce a concept for generics and a `<script>` attribute for marking a component as having no other events besides the ones defined within.
 
 While this is not a change to Svelte's core, it's still something that needs to be specified so intellisense implementers have something to adhere to.
 
@@ -66,14 +66,14 @@ Now you add one event which comes from a dispatcher mixin:
 <button on:click>Forwarded</button>
 ```
 
-In this case `strictEvents` will not work anymore because we cannot know that `mixinDispatch` dispatches events. So now you use the `ComponentEvents` interface.
+In this case `strictEvents` will not work anymore because we cannot know that `mixinDispatch` dispatches events. So now you use the `$$Events` interface.
 
 ```html
 <script lang="ts">
     import {mixinDispatch} from "somewhere";
     import {createEventDispatcher} from "svelte";
 
-    interface ComponentEvents {
+    interface $$Events {
         mixinEvent: CustomEvent<string>;
         own: CustomEvent<boolean>;
         click: MouseEvent;
@@ -93,7 +93,7 @@ This works the same as for typing events.
 
 ```html
 <script lang="ts">
-    interface ComponentSlots {
+    interface $$Slots {
         default: { prop: boolean; };
     }
 </script>
@@ -107,7 +107,7 @@ This works the same as for typing events. You probably won't use that because it
 
 ```html
 <script lang="ts">
-    interface ComponentProps {
+    interface $$Props {
         prop: boolean;
     }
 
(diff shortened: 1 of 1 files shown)


numfin:
Why not use 1 interface for props/events/slots ? This way we can create events, that depends on props:
```ts
interface ComponentDefinition<T extends Record<string, string>> {
  props: { a: T },
  events: { b: T }
}

```

p.s. Sorry for bad english, tried my best.

dummdidumm:
Yes this would be possible through the [ComponentDef interface](https://github.com/dummdidumm/rfcs/blob/ts-typedefs-within-svelte-components/text/ts-typing-props-slots-events.md#componentdef)

stefan-wullems:
For the slots, props and events, I would lose the `Component` prefix: 
 - `ComponentSlots` -> `Slots`
 - `ComponentProps` -> `Props`
 - `ComponentEvents` -> `Events`

Then finally `ComponentDef` -> `Component`.

Doesn't make the names a lot more ambiguous or prone to conflict with existing types I think.

stefan-wullems:
Perhaps separating the type definition from the actual logic of a component would work nicely.

```tsx
<script lang="ts" definition>
  interface Component<T> {
     props: { value: T }
     events: { 
       change: (value: T) => void
     }
     slots: {
       default: {}
     }
  }
</script>

<script lang="ts">
  export let value
</script>

<some>
  <slot>Markup<\slot>
</some>
```

dummdidumm:
I like the shortening of the names although I think this might increase the possibility of name collisions.
I'm against putting this into a separate script. This would require more effort in preprocessing, also I like the colocation of the interfaces within the instance script.

AlexGalays:
Can't wait to use this <3 

- Basic components without generics means mapping domain models from/to arbitrary component specific structures.
- Passing a component to a slot is impossible so our best workaround for now would be to pass a [Component, ComponentProps] tuple but there's no way to have ComponentProps mean anything right now.

🙏

dummdidumm:
Could you elaborate on your second point a little more? I'm don't fully understand the use case. And what do you mean by "impossible"? Possible to do in Svelte but the type checker complains?

AlexGalays:
(unless I missed something)

you can't pass a component instance to a slot so people end up either 
- wrapping all their slots with a div on the call site, ending up in a div soup or
- passing a component and its props as a non slot prop:  `<ParentComponent renderHeader={[SomeSvelteComponent, Props]} />`

but today, we have no way to express that `Props` should be SomeSvelteComponent's Props beside triple checking manually.

DboyRedd (approved):

francoislg:
Stumbled upon this and just wanted to throw here a slight variation of Option 2:

```
<script lang="ts">
    import {createEventDispatcher} from "svelte";

    type T = ComponentGeneric<boolean>; // extends boolean
    type X = ComponentGeneric; // any

    export let array1: T[];
    export let item1: T;
    export let array2: X[];
    const dispatch = createEventDispatcher<{arrayItemClick: X}>();
</script>
```

I think it would be slightly closer to TypeScript code than a `ComponentGenerics` interface that gets magically expanded :smile:

denis-mludek:
Hi !

Any idea when Generic Component will be available/released? Is it perhaps a case of choice paralysis? I think we would all love something even if it's 100% perfect!!

I'm working on a SvelteTS project for several weeks now, and I would have used this feature a few times already.
Btw, love the work you're doing to make SvelteTS a thing. TS support was the thing that made me switch from React to Svelte. 🤗

dummdidumm:
@tomblachut tagging you since you are the maintainer of the IntelliJ Svelte Plugin - anything in that proposal that concerns you implementation-wise ("not possible to implement on our end")?

tomblachut:
@dummdidumm thank you for tagging me.

Generics will definitely be, as you've written, an uncanny valley and maintenance burden.

Option 1 & 2 without Svelte support in editor will produce invalid TS, given that both will require special reference resolution.  I think it's better to avoid that.

I'd scratch Option 2, because ComponentGenerics is in the same scope as things that will refer to its type parameters. I imagine it will add some implementation complexity AND mental overhead for users.

I quite like Option 3 because it's valid TS. ComponentGeneric would be treated as identity mapped type.  
```ts
    type T = ComponentGeneric<boolean>; // extends boolean
    type X = ComponentGeneric; // any
``` 

Option 3 could be even simplified a bit by giving new semantics to `export type` in similar way as `export let` denotes a prop
```ts
    export type T = boolean;
    export type X = any;
``` 

Now, I think it's better to stick to one style of declarations: (separate interfaces/compound ComponentDef/namespace) otherwise we may introduce small bugs in one of them and more importantly will need to decide on and teach about precedence.

One additional thing this proposal does not mention is ability to extend interfaces. I think that's great feature. Author of the component could say "this "PromotedPost adheres to Post props" and whenever types are changed in Post definition, implementing components would show type errors. Unless I'm missing something interfaces will support that use case out of the box.

dummdidumm:
Thanks for your insights! I agree that option 3 feels the closest to vanilla TypeScript which is why I prefer that, too. I also agree that we should only provide one style of declarations. Separate interfaces feels like the best option there.

One thing you brought up about extending interfaces is a very good advantage, but it also got me thinking how to deal with generics in that context.

For example you have this interface:

```ts
export interface ListProps<ListElement> {
   list: ListElement[];
}
```

How do you extend it while keeping it generic in the context of a Svelte component? The only possibility that comes to my mind is to do this:

```svelte
<script lang="ts">
  import type { ListProps } from '..';

  type BooleanListElement = ComponentGeneric<boolean>;
  interface ComponentProps extends ListProps<BooleanListElement> {}

  export let list: BooleanListElement[];
</script>
..
```

non25:
I miss generics too.
I'll leave one example which I would really appreciate to use generics with:

This checkbox selector returns subset of an array of objects without mutating them, which is really handy, but it will return `any[]` instead of passing types forward, which is sadge... :disappointed: 

```svelte
<script>
  import Checkbox from '../components/Checkbox.svelte';

  export let checkboxes = [];
  export let checked = [];
  export let idF = 'id';
  export let textF = 'text';

  let checkedIds = new Set(checked.map((c) => c[idF]));

  function mark(id) {
    checkedIds = new Set(
      checkedIds.has(id)
        ? [...checkedIds].filter((cid) => cid !== id)
        : [...checkedIds, id]
    );
  }

  $: checked = checkboxes.filter((c) => checkedIds.has(c[idF]));
</script>

<ul class="checkboxes">
  {#each checkboxes as checkbox (checkbox[idF])}
    <li>
      <Checkbox
        checked={checkedIds.has(checkbox[idF])}
        on:change={() => mark(checkbox[idF])}
        desc={checkbox[textF]}
      />
    </li>
  {/each}
</ul>
```

This is proposed way to use generics? I don't think I understood examples correctly, so I added type definitions to this component.

```svelte
<script lang="ts">
  import Checkbox from '../components/Checkbox.svelte';

  type Item = ComponentGeneric; // how this will attach to the checkboxes prop ?

  export let checkboxes = Item[];
  export let checked = Item[];
  export let idF = 'id';
  export let textF = 'text';

  let checkedIds = new Set<number>(checked.map((c: Item) => c[idF]));

  function mark(id) {
    checkedIds = new Set(
      checkedIds.has(id)
        ? [...checkedIds].filter((cid: number) => cid !== id)
        : [...checkedIds, id]
    ):
  }

  $: checked = checkboxes.filter((c: Item) => checkedIds.has(c[idF]));
</script>
```

dummdidumm:
Everything inside this proposal is type-only, which means it's only there to assist you at compile time to find errors early - nothing of this proposal will be usable at runtime, similar to how all TypeScript types will be gone at runtime.

In your example you would do:

```svelte
<script lang="ts">
  // ..
  type Item = ComponentGeneric;
  type ItemKey = ComponentGeneric<keyof Item>;

  export let checkboxes: Item[] = [];
  export let checked: Item[] = [];
  // Note: For the following props, the default value is left out because you cannot expect "id" or "text" to be present as a default if you don't narrow the Item type
  export let idF: ItemKey;
  export let textF: ItemKey;
// ..
```

non25:
Wow, keyof is cool.

I want to make sure that I understand correctly, so here's another example:

```svelte
<script lang="ts">
  interface Vegetables {
    id: number;
    name: string;
    weight: number;
  }

  let someItems: Vegetables[]: = [
    ...
  ];

  let checked = Vegetables[];
</script>

<CheckboxSelector checkboxes={someItems} textF="name" idF="id" bind:checked />
<!--                            ^                ^ 
                                |                | Could it attach to this?
                                |
                                | how do I specify that ComponentGeneric
                                  should attach to this?
-->
```

For example here's how I would use this in regular typescript, which is clear to me.

```ts
function component<T>(checkboxes: T = [], idF: <keyof T>, textF: <keyof T>) {
  ...
}
```

dummdidumm:
When using the component, you don't specify anything, you just use the component and the types should be inferred and errors should be thrown if the relationship between is incorrect. So in your example if you do `textF="nope"` it would error that `nope` is not a key of `Vegetables`.

Doing 
```ts
type T = ComponentGeneric;
export let checked: T[];
export let idF: keyof T;
// ...
```
Would be in the context of Svelte components semantically the same as
```ts
function component<T>(checked: T[], idF: keyof T) {
  ...
}
```
This is the uncanny valley I'm talking about in the RFC which I fear is unavoidable - it doesn't feel the same like generics, yet it serves this purpose inside Svelte components (and inside only Svelte components). The problem is that Svelte's nice syntax of just directly starting with the implementation without needing something like a wrapping function, so there is no place to nicely attach generics.

non25:
So if I end up in a situation where I need two generics:

```typescript
function component<X, Y>(items: X[], someOtherProp: Y) {
  ...
}
```

How that would look in the proposed approach? :thinking: 
Is it even possible ?

Do you know how bindings will work in the current state?

In the example above I added annotation that `checked` in parent component is of same type as `someItems`, but if we pass `someItems` through `checkboxes` prop, we should lose the type, because we can only use any as 'generic' type now in the `CheckboxSelector` component, right ?

Do annotations to a bind override `any[]` in this case ? :thinking: 

Sorry for the wording...

dummdidumm:
> 
> 
> So if I end up in a situation where I need two generics:
> 
> ```ts
> function component<X, Y>(items: X[], someOtherProp: Y) {
>   ...
> }
> ```
> 
> How that would look in the proposed approach? 🤔
> Is it even possible ?
> 

You do this
```ts
type X = ComponentGeneric;
type Y = ComponentGeneric;
export let items: X[];
export let someOtherProp: Y;
```

> Do you know how bindings will work in the current state?
> 
> In the example above I added annotation that `checked` in parent component is of same type as `someItems`, but if we pass `someItems` through `checkboxes` prop, we should lose the type, because we can only use any as 'generic' type now in the `CheckboxSelector` component, right ?
> 
> Do annotations to a bind override `any[]` in this case ? 🤔
> 
> Sorry for the wording...

Sorry, I don't know what you mean.

non25:
> Sorry, I don't know what you mean.

I hope this is a better way to explain. :thinking: 

```svelte
<script lang="ts">
  interface Vegetables {
    id: number;
    name: string;
    weight: number;
  }

  let someItems: Vegetables[]: = [
    ...
  ];

  // I annotate same type to the checked prop, which I will bind below
  // What type checked will have after the bind ?
  let checked = Vegetables[];
  // I need to use checked in other places and want it to retain the type
</script>

<CheckboxSelector checkboxes={someItems} textF="name" idF="id" bind:checked />
<!--                                                                  ^
                                                          binding checked here
-->

<script lang="ts">
  // ...
  // I set any here because I want to accept any object type
  // and generics is currently not supported
  export let checkboxes: any[] = [];

  // this prop will contain a subset of checkboxes prop,
  // which I bind above
  export let checked: any[] = [];
  // ...
</script>
```

dummdidumm:
To me this sounds like you mix up some of TypeScript's type system with regular JavaScript. For example `let checked = Vegetables[]` is not valid TypeScript, because `Vegetables` is an interface, which does not exist at runtime. It should be `let checked: Vegetable[]`, which also answers your second question: The type will stay `Vegetable` because you told TypeScript that it's of this type, that does not change. Instead, TypeScript would a compiler error if you would try to assign something to `checked` that does not suffice the `Vegetable` interface (`any` is okay, because you can assign `any` to anything).

non25:
> To me this sounds like you mix up some of TypeScript's type system with regular JavaScript.

I'm just making mistakes, not used to type `let something: type[] = [];` :grin:
Thanks for the explanation. That's better than nothing.

So looks like the most negatively affected use cases without generics are:

- passing types down through some wrapper component like `svelte-viewpoint`
- slot let templates for generic arrays

?

tomblachut:
Generics come in handy if you want connect types of 2 or more things to each other. Fo example you may want to guarantee that some event's payload will be of the same type as a prop, or some derivative of that type.

tomblachut:
@dummdidumm 
> Option 3 has one little shortcoming though, being not being strict enough.

Ouch, TIL. Thanks for catching that. Actually not "TIL", I spend couple of days pondering about that 😅 I agree that it's a hit we can take since this problem has additive solution. We can wire up annotations for this on top of normal TS, in contrast to suppressing some false-negatives. 👍 

`export` has one advantage over `ComponentGeneric` - it's not a reference. Though, if it's okay to enable `import { ComponentGeneric } from 'svelte';` I'm okay with both approaches. I'd just prefer to avoid global implicit type.

---

```svelte
<script lang="ts">
  import type { ListProps } from '..';

  type BooleanListElement = ComponentGeneric<boolean>;
  interface ComponentProps extends ListProps<BooleanListElement> {}

  export let list: BooleanListElement[];
</script>
```
I think your example is elegant.

I didn't consider yet how extends will integrate with `SvelteComponentTyped`. Those 3 interfaces would need to be translated into generics of that class. Will it even work?

dummdidumm:
Implicit global type - I'm a little split on this. There are some other globals already after all, on the other side it's more explicit. Either way, its presence would need extra tooling code (that "it's not the same in TS" thing I mentioned earlier).

About adding it to `SvelteComponentTyped`: Do you mean from a tooling perspective how to automatically generate type definitions for libraries, or how to generate the code for the IDE? Either way, some additional logic would be necessary to first extract the interface out to the top level so they can be used on the class, and then another step where the `ComponentGeneric` usages are collected and added in order to `SvelteComponentTyped` like `export class MyComponent<T, X extends string> extends SvelteComponentTyped<{ prop: T },..`. Interfaces could be extracted the same, so that 
```ts
type T = ComponentGeneric;
interface ComponentEvents {
  foo: CustomEvent<T>
}
```
Becomes
```ts
interface ComponentEvents<T> ...

export class MyComponent<T> extends <.. ComponentEventd<T>..
```

pitaj:
I understand wanting to use as little extra grammar as possible, but I think we should consider using `export` to signify a generic type and for specifying the interface of props, slots, and events.

This has a few benefits:

- avoids the use of magic reserved types
- allows for default types
- fully backwards compatible, since it is currently invalid syntax
- use shorter name for the interfaces for slots, props, and events

```svelte
<script lang="ts">
    import {createEventDispatcher} from "svelte";

    // required generic type extending boolean
    export type T extends boolean;
    // optional generic type
    // defaults to string
    export type X = string;
    
    // instead of ComponentSlots
    export interface Slots {
        default: { aSlot: T }
    }

    export let array1: T[];
    export let item1: T;
    export let array2: X[];
    const dispatch = createEventDispatcher<{arrayItemClick: X}>();
</script>
```

The only thing in the example that isn't valid typescript syntax is `export type T extends boolean`, which IMO is not asking for much.

dummdidumm:
Adding invalid TS syntax is a no-go for me, it would require every IDE to know of this and people need to get used to that for Svelte only.
About the exports: I'm hesitant because exports are right now reserved for props. On the module level, you can also export interfaces/types, but the meaning is different. I'm not sure if it would bring more confusion than helping. On the other hand one could argue that it does align with the "public API" semantics of `export let`.

If we go with `ComponentGeneric` I think I have one more argument in favor of implicit global: If it's an explicit import, people might think they can model their API like this outside of Svelte components, which is wrong. With a global type, they would see that TS will throw an error ("unknown type") if you try that outside.

Monkatraz:
Just throwing stuff at the wall...
```html
<script lang="ts">
  import {createEventDispatcher} from "svelte";

  // this is wacky as all hell but kinda sorta less globals (syntax is more rigid??)
  type T<V extends boolean = Generic> = V
  type X<V = Generic> = string

  export let array1: T[];
  export let item1: T;
  export let array2: X[];
  const dispatch = createEventDispatcher<{arrayItemClick: X}>();
</script>
```
Honestly, this is the only alternative I could come up with. I think `ComponentGeneric` is probably the way to go unless anyone gets any ideas or an epiphany. Maybe my attempt will inspire someone, lol.

Monkatraz:
Also, wouldn't using `$$` prefixes instead of `component` match Svelte patterns better?
```html
<script lang="ts">
    import {createEventDispatcher} from "svelte";

    type T = $$Generic<boolean>; // extends boolean
    type X = $$Generic; // any
    
    // you can use generics inside the other interfaces
    interface $$Slots {
        default: { aSlot: T }
    }

    export let array1: T[];
    export let item1: T;
    export let array2: X[];
    const dispatch = createEventDispatcher<{arrayItemClick: X}>();
</script>
```

pitaj:
> Adding invalid TS syntax is a no-go for me, it would require every IDE to know of this and people need to get used to that for Svelte only. Moreover, there wouldn't be things like optional generics. Generics need to be driven by the prop input.

Can you explain more why optional generics shouldn't be a thing? I'm not quite grasping your reasoning.

But if there aren't optional types, you can just use the right hand side as the base type. 

```
type T = string; // generic type T extends string
```

> I'm hesitant because exports are right now reserved for props. On the module level, you can also export interfaces/types, but the meaning is different. I'm not sure if it would bring more confusion than helping. On the other hand one could argue that it does align with the "public API" semantics of export let.

They're modeled explicitly after the way props work. I think following that model would make following this more intuitive and more discoverable. For instance, setting a generic could be just like using a prop:

```
// Foo.svelte
<script lang="ts">
type X = string;
</script>

// Bar.svelte
<script lang="ts">
import Foo from './foo';

interface Stuff {}
</script>

<Foo X={Stuff} />
```

I don't think setting them explicitly is planned as part of this rfc it's just an example to demonstrate my line of thinking.

Before I knew this wasn't possible yet, I tried doing stuff similar to this (to be met with errors of course).

tomblachut:
@pitaj Sorry, but we can't make following happen:
```
<Foo X={Stuff} />
```
You can already create an uppercase prop (https://svelte.dev/repl/2897e47e9abf4858adc46978db9a1717?version=3.33.0), passing generics explicitly would require different syntax. This proposal doesn't cover that, nor should it. Let's agree on first part that is generics declaration + possibility to infer them from props/events/slots passed to the component.


```
export type T extends boolean;
```
Strong no from me. This is invalid TS, we can't invent new syntax, in addition to @dummdidumm's points, what if TS decides to add `extends X` part in future version with different meaning that ours.

---

> If we go with ComponentGeneric I think I have one more argument in favor of implicit global: If it's an explicit import, people might think they can model their API like this outside of Svelte components, which is wrong. With a global type, they would see that TS will throw an error ("unknown type") if you try that outside.

That's a fair point. I still prefer `export` but you're dragging me to your side step by step. I thought a bit more about implicit global implementation and it should be easier than I initially anticipated.  This would be slightly parallel to `$$props` so maybe @Monkatraz`s  `$$Generic` is a better fit.

pitaj:
> Sorry, but we can't make following happen

I wasn't suggesting it, I was just demonstrating how my thought process went while initially looking for this feature.

> Strong no from me. This is invalid TS, we can't invent new syntax

That's fair, alternatives have been proposed.

dummdidumm:
I agree with the `$$`-syntax, it chimes well with the Svelte "language". Updates the rfc accordingly.

We did discuss this in the maintainer's meeting and I got positive feedback on the current state. One thing that came up is the possibility for the ability to export the interfaces for usage in other components to extend it there. Got to think about how that might work in combination with generics (maybe it won't).

pitaj:
Does this rfc desire not to cover explicitly designating generics, as in, generics would only be possible for cases where their type can be derived from the props given?

For explicitly designating types, would it not be possible to add a new syntax similar to a prop modifier `generic:<GenericTypeName>(={<Type>})` that could be used like so:
```html
<Component generic:X={string} />
```

dummdidumm:
Explicitly declaring the generic is out of scope of this rfc, and at the moment I'm not even sure something like this is needed, since I don't see how generics can be driven by something else than props - and if you pass in these props, the generic type should be inferred from that.

pitaj:
I think one is where the prop types are derived from the generic:

types.d.ts
```ts
interface A<T> { ... }
interface B<T> { ... }
```
Component.svelte
```html
<script type="ts">
type X = $$Generic;

export let a: A<X>;
export let b: B<X>;
</script>
```
Other.svelte
```html
<script type="ts">
import Component from './Component.svelte'

type Union = string | number;

let a: A<Union> = 'Hello, world';
let b: B<Union> = 123;
</script>

<Component a b />
```
Inference probably could figure out that `X == Union`, but does this hold in any case, no matter the complexity?

gustavopch:
I know explicitly passing type parameters is out of the scope, but I'd just like to point that it's an already resolved thing in JSX, so if it's ever needed for Svelte, JSX can be followed:
```jsx
<Formik<FormValues>
  initialValues={initialValues}
  validateOnBlur={false}
  validate={onValidate}
  onSubmit={onSubmit}
>
```

dummdidumm:
The Svelte compiler would need to be adjusted then, which would be the first time that would need to be done for a typescript feature.

AlexGalays:
> I know explicitly passing type parameters is out of the scope, but I'd just like to point that it's an already resolved thing in JSX, so if it's ever needed for Svelte in the future, JSX can be followed:
> 
> ```js
> <Formik<FormValues>
>   initialValues={initialValues}
>   validateOnBlur={false}
>   validate={onValidate}
>   onSubmit={onSubmit}
> >
> ```

That doesn't do a whole lot though? Beside more precise compilation errors perhaps. Nowadays TS is pretty good at inferring so as long as each individual props inside the component are well typed relative to each other, it seems redundant to explicitely set the type on the call site? Anyway, it's not a must-have feature like the one discussed here!

gustavopch:
@AlexGalays I'm not using Svelte in my projects yet, so I can't really say how useful explicit type parameters would be. What I do know is that in JSX they are very useful for example in the following snippet:
```jsx
<Formik<FormValues>
  initialValues={{
    name: ''
  }}
/>
```
Without an explicit type parameter, `initialValues` would accept anything and try to infer the type from that. But what if I already have a type and want to make sure `initialValues` conforms to it? Then I have to either (a) declare `const initialValues: FormValues = { ... }` somewhere outside of the JSX or (b) pass `FormValues` as an explicit type parameter and keep `initialValues` inlined.

Anyway, I'm not saying that it should be implemented, I'm just throwing some information for someone reading this in the future **in case** it's ever decided explicit type parameters are needed.

dummdidumm:
Experimental support for typing props/events/slots and generics is now available as described in this RFC. Please provide feedback in the following issue: https://github.com/sveltejs/language-tools/issues/442
The feature is experimental and subject to change. It therefore doesn't follow semantic versioning yet.

derolf:
Is it possible to narrow the type of $$Generic? Like:

```
type ITEM = $$Generic & SomeBaseType;
```

So to enforce that the passed type extends `SomeBaseType`?

dummdidumm:
https://github.com/sveltejs/language-tools/issues/1326 proposed to add style props (`<Component --style-prop="value" />`) to this, too. It also brought up the good question about "what do JS users do?" because they can't use interfaces, they have to rely on JSDoc. We should probably enhance the RFC and find a JSDoc-compatible version, too. I think we can take inspiration for the API from the great work of [sveld](https://github.com/carbon-design-system/sveld#api-reference) (kudos to @metonym).

Note to self: If we add support for generics via JSDoc, we need to adjust the transformation because right now it relies on using TS syntax in the transform output (since we know it's TypeScript). Not sure if we can use the same transformation then.

VdustR:
Thank you for the awesome work!

Is there example for generic parameter defaults? I can't find that in the RFC.

```ts
function foo<T extends Item = string>(..)
```

<https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html#generic-parameter-defaults>

dummdidumm:
What's the use case for specifying defaults? Could you give an example where this is helpful/needed?

VdustR:
I made a function with generic parameter defaults like this:

```ts
export const defaultGen = () => Math.random().toString(16).substring(2);

function Likftc<
  Source extends number | string,
  Target extends number | string = string
>(
  initialKeys: Source[] = [],
  generator: () => Target = defaultGen as () => Target
)
```

Full code: <https://github.com/VdustR/likftc/blob/f0203a9d3e71a9513d396f4809369778a3dfbc25/packages/core/index.ts#L3>

The function type will be like:

```ts
let target = "foo";
Likftc([1, 2, 3], () => target).get(1) // get: (item: 1 | 2 | 3) => string

let target = 1234;
Likftc([1, 2, 3], () => target).get(1) // get: (item: 1 | 2 | 3) => number

Likftc([1, 2, 3]).get(1) // get: (item: 1 | 2 | 3) => string

// without generic parameter defaults
Likftc([1, 2, 3]).get(); // get: (item: 1 | 2 | 3) => string | number
```

With the generic parameter defaults I can use the default type if the argument is not passed.

I also make a react hook depending on it like this:

```ts
function useLikftc<
  Source extends number | string,
  Target extends number | string = string
>(frame: Source[], generator?: () => Target)
```

Full code: <https://github.com/VdustR/likftc/blob/f0203a9d3e71a9513d396f4809369778a3dfbc25/packages/react/index.ts#L4>

And it will work like:

```ts
let target = 123;
useLikftc([1, 2, 3], () => target).get(); // get: (item: 1 | 2 | 3) => number

let target = "foo";
useLikftc([1, 2, 3], () => target).get(); // get: (item: 1 | 2 | 3) => string

useLikftc([1, 2, 3]).get(); // get: (item: 1 | 2 | 3) => string

// without generic parameter defaults
useLikftc([1, 2, 3]).get(); // get: (item: 1 | 2 | 3) => string | number
```

I use component with `let:` directive as the react custom hook alternative in svelte (reference: <https://gradientdescent.de/custom-hooks/#Mypreferredsolution>). The svelte component props are like this:

```ts
type Source = $$Generic<number | string>;
type Target = $$Generic<number | string>; // here is the generic parameter defaults needed
export let keys: Source[];
export let generator: (() => Target) | undefined = undefined;
```

Full code: <https://github.com/VdustR/likftc/blob/f0203a9d3e71a9513d396f4809369778a3dfbc25/packages/svelte/index.svelte#L8>

The component would be used like:

```svelte
<Likftc keys={[1, 2, 3]} generator={() => 'foo'} let:get />  <!-- get: (item: number) => string -->
<Likftc keys={[1, 2, 3]} generator={() => 'foo'} let:get />  <!-- get: (item: number) => number -->

<Likftc keys={[1, 2, 3]} let:get />  <!-- get: (item: number) => string | number -->
         <!-- It's better to be typed as `get: (item: number) => string` in this case  -->
```

With the generic parameter defaults, the component could be defined with the default type if the property is not passed. For now `let:get` will return with type `get: (item: number) => string | number` if I don't pass `generator` and it's better to be typed as `get: (item: number) => string`.

icalvin102:
I just stumble across this the other day and I really like it. Thanks for the awesome work.

One thing I noticed though: If I try to override a event that is already predefined by svelte it wont give the correct type.

```ts
const dispatch = createEventDispatcher<{'input': Foo, 'customInput': Foo }>();
```
Produces the following types:
```ts
on:input // CustomEvent<any>
on:customInput // CustomEvent<Foo>
```

This is true for all predefined Events that I tested (`submit`, `change`, `input` ...)

The ability to override would be very nice IMO.

hperrin:
Is there a way to define defaults on generics? In TS, you can do `T extends SomeInterface = SomeClass`, and it will know, if you didn't supply T, that T is the `SomeClass` type. Can we do the same with this solution?

hperrin:
I also seem to be having trouble returning a generic from a function. No matter what you give the generic, it always types the return value in the widest way possible. Here's an example:

```svelte
 <!-- Example.svelte -->
<svelte:element this={tag} bind:this={element} {...$$restProps}
  ><slot /></svelte:element
>

<script lang="ts">
  type TagName = $$Generic<'input' | 'span'>;
  interface $$Props
    extends svelte.JSX.HTMLAttributes<HTMLElementTagNameMap[TagName]> {
    tag: TagName;
  }

  export let tag: TagName;

  let element: HTMLElementTagNameMap[TagName];

  export function getElement() {
    return element;
  }
</script>
```

```svelte
<!-- UseExample.svelte -->
<Example tag="input" bind:this={example} />

<script lang="ts">
  import { onMount } from 'svelte';
  import Example from './Example.svelte';

  let example: Example<'input'>;

  onMount(() => {
    const el = example.getElement();
  });
</script>
```

When you hover over `el`, you get `const el: HTMLInputElement | HTMLSpanElement` as the type, and trying to use `el.value` results in:

```
Property 'value' does not exist on type 'HTMLInputElement | HTMLSpanElement'.
  Property 'value' does not exist on type 'HTMLSpanElement'.ts(2339)
```

Am I correctly following the RFC in this example? Is this an expected error?

AlbertMarashi:
I'm getting a bug when nesting multiple generic types within components

`Main.svelte`
```html
<script>
import Foo from "./Foo.svelte"

</script>
<Foo xyz="foo" >
    <div slot="option" let:xyz>
        <h1>{xyz}</h1>
    </div>
</Foo>
```
The type of xyz here is `unknown` when it should be `string`

`Foo.svelte`
```html
<script lang="ts">
import Bar from "./Bar.svelte"

type T = $$Generic
export let xyz: T

</script>
<Bar {xyz}>
    <svelte:fragment slot="option" let:bar>
        <slot name="option" xyz={bar}/>
    </svelte:fragment>
</Bar>
```

`Bar.svelte`
```html
<script lang="ts">
type T = $$Generic
export let xyz: T

</script>
<slot name="option" bar={xyz} />
```

Tal500:
I've tried the generic feature of components, really helpful!
Suggestions:
1. Like it was told at least twice here, I think it should support default generic arguments. My use case is when you want to get the component as a prop of a different component (when you want to save the result of the component via `bind:this`, or when you want to pass a component constructor to a different component). Yes, I could say explicitly what I want there, but why? Having a default value also suggests the user (that use a generic component library, say) what the 'default-behavior' or the 'unspecified' value should be.
2. We can allow typing on JS (including the generics), if we will use JSDoc to embed TS-defs. For example, we can define the generics in JS via [`@typedef`](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#typedef-callback-and-param).

Tal500:
Another question I hope is related here - how to get the self component type? It is useful in my cases to pass it to generics, but can also be helpful for recursion.
Suggestion - `$$Self`.

Tal500:
Another feature that was "nice to have" before generics, but now it become more crucial:
Allow the user to infer the component type when binding the component to variable, by having a special meta-type called `$$Infer`, e.g.:
```svelte
<script lang='ts'>
  import MyComp from './MyComp.svelte';

  let comp: $$Infer;// $$Infer should be equals here to the type `MyComp`;
</script>

<MyComp bind:this={comp} />
```

You may look at the example I gave and think for yourself - the user can simply type the var `comp` as `MyComp` by himself, why do we need to introduce the new language-feature `$$Infer` keyword?

So the reason is for the case that the component designer want to use generics, sometimes in sophisticated ways (e.g. me :-) ), and want to infer the generic type argument by the types of values the user gives as properties to the component.
At first I thought to myself - yey! auto-type-inference of properties! But then quickly after I realize that when I want to bind this component to a variable, I need to work hard on specifying the correct generic type arguments, since inference doesn't work on the variable I'm binding the component to, unless we take my suggestion.

Why I chose the keyword of `$$Infer`? Because it somewhat reminding the type script syntax of [`T extends S<infer R> ? A : B`](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#inferring-within-conditional-types).

We can also take this suggestion a step further, and say we sometimes want the user to force a certain type of component, but give him the ability to let the compiler to deduce the generic arguments by itself, e.g.:
```svelte
<script lang='ts'>
  import MyCompGeneric from './MyCompGeneric.svelte';
  import MyOtherComp from './MyOtherComp.svelte';

  let num: number = 3;

  let comp: MyCompGeneric<$$Infer>;// $$Infer should be equals here to the type `{ a: number; }`
  // If we were replacing the previous line with the following, there should be a typing error:
  //let comp: MyOtherComp;
</script>

<MyCompGeneric bind:this={comp} propA={ { a: num; } } />
```

AlbertMarashi:
https://github.com/sveltejs/rfcs/pull/38#issuecomment-1310036123

This would be great

marekdedic:
Hi,
what is the state of this RFC? I see it has been implemented for a year, however, it is still referred to as experimental.

I read the whole discussion here and it the feedback issue and it seems to me that while there is a lot of discussion about generics, the rest of the RFC seems to be pretty non-controversial. So, could we get things like `strictEvents`, `$$Events`, `$$Props` and `$$Slots` stabilized?

Thanks!

wvhulle:
Hi it seems like this is already working on the latest releases but I couldn't find a reference to this in the docs. Is there a reason it is left out?

Tal500:
> Hi it seems like this is already working on the latest releases but I couldn't find a reference to this in the docs. Is there a reason it is left out?

I guess the reason is that they want to make sure they wouldn't regret the final API (i.e. commiting to the API, API stability), and since they're focusing on the other Svelte products now.

dummdidumm:
Exactly, this API is not finalized yet - it's experimental and might still change, so we don't want docs for it yet.

ivanhofer:
@wvhulle If you are looking for some guidance on how to implement something you can [go to this repository](https://github.com/ivanhofer/sveltekit-typescript-showcase) where I documented all the things I find useful when working with `TypeScript` inside `Svelte` appplications.

dummdidumm:
A lot of time has passed since opening this RFC, with plenty of time to try it out.

One thing I'm not so sure anymore is the decision about [generics](https://github.com/dummdidumm/rfcs/blob/ts-typedefs-within-svelte-components/text/ts-typing-props-slots-events.md#generics). Currently you do a generic like this:

```svelte
<script lang="ts">
  type T = $$Generic;
  type X = $$Generic<string>;

  export let items: T[];
  export let key: X;
</script>
```

This requires the reader to know about the special generics implementation of Svelte, and then map it in your head to the corresponding TypeScript syntax.

A initially discarded alternative would be to use a generics attribute, like this:

```svelte
<script lang="ts" generics"T, X extends string">
  export let items: T[];
  export let key: X;
</script>
```

It's a bit weird that generics are not inside the script tag but on its attribute, but the advantage is that it reads a lot closer to the TypeScript equivalent.

I'd like to have everyone's thoughts on which solution they would prefer. (Please vote with thumbs up/down and if you want to, additionally leave a comment)

marekdedic:
Hi,
IMHO the `$$Generic` proposal seems better because it is just TypeScript with a weird type added - however, things like editor integration, third-party tool support (prettier, eslint) and syntax highlighting would still work with this. the prop approach has none of these - the generics definition is just a string and I think all the tools I mentioned would have a much harder time working with it...

With that said, if there is no consensus around this, could I propose stabilizing the rest of this RFC (`strictEvents`, typed slots, props and events) first?

ivanhofer:
I also see the `generics` prop on the script tag problematic becasue of the missing tooling. Good tooling for this needs to be built and maintained. Leaving the `$$Generic` works already with the existing tooling.

And having `$$Props` inside the script tag but `$$Generic` not feels inconsistent. Just because `vue` now uses this syntax, doesn't mean `Svelte` should use it too. Exposing just `$$Genreic` on the script tag, but leaving everything else untouched feels wrong. I would leave it as it is or come up with a new approach that touches all types affecting a component.

dummdidumm:
Oh, interesting, I didn't know Vue has added the `generics` attribute syntax - if the decision is close, this could be a factor (transferability of knowledge).

Regarding tooling: What missing tooling are you talking about exactly? language-tools would of course implement this if it was decided to go that way, which would entail the eslint plugin and the language server which powers the VS Code extension.

ivanhofer:
@dummdidumm I saw it on twitter the other day: https://twitter.com/youyuxi/status/1639919409312710659

Agree, the transferability of knowledge would be the only reason I currently see as a benefit.

It somehow feels weird to me to define part of the types on the script tag and others like `$$Props` and `$$Events` inside the script tag.

**Tooling:**
well everything that falls into it when using generics in regular `TypeScript`
 - syntax errors
 - code completion
 - ...
image

Without tooling this would be a simple string where you can type anything.

I don't know the implementation details but it seems a bit of an effort to get everything running.

What about syntax changes? Recently the [`const Type Parameters`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#const-type-parameters) were added. Will this work out of the box if someone updates their `TypeScript` dependency in their project?

Kind of related to https://github.com/sveltejs/svelte/issues/4701, where the tooling would need to be injected into a small portion of other syntax.

marekdedic:
Ok, then maybe this is a bit brazen, but let me tag @yyx990803 here so that they see the discussion here and maybe have something to add...

By tooling I meant things such as [eslint-plugin-svelte](https://github.com/ota-meshi/eslint-plugin-svelte) - yes, the special attribute syntax could of course be implemented there, but the less "magic" there is, the better chance any 3rd party tools designed for TS would work with Svelte as well...

dominikg:
even if the script tag attribute feels appealing syntax wise, the amount of extra work it imposes on tooling is too large and it's something new we have to teach/document. And it's kind of tucked away, so readability is less prominent.

dummdidumm:
What kind of extra work does this impose on tooling compared to the `$$Generic` solution? Both solutions need special treatment by tooling or else they produce semantical invalid code:
- `type X = $$Generic` needs at least a injected `$$Generic` global so that it does not error ("`$$Generic` is not defined"). In order to understand that this is in fact a generic it needs elaborate transformations under the hood
- `generics="X"` needs at least an injected `X` type to that it does not error on usage locations ("`X` is not defined"). In order to understand that this is in fact a generic it also needs transformations under the hood

Regarding possible tooling:
- AFAIK Intellij supports neither today, but if Vue now adopted the `generics` syntax we might have better chances of getting support for this in Svelte as they could hopefully reuse some of their solution for Vue
- `eslint-plugin-svelte` (the designated new official eslint plugin, soon) supports neither today
- `prettier-plugin-svelte` does not have to be aware of this
- language tools (VS Code, `svelte-check` etc) support `$$Generics` today. I think supporting `generics` will be roughly the same amount of complexity

Assuming both solutions require the same amount of tooling efforts (in other words, assume this is not a factor) - what would be your preference?

dummdidumm:
With the addition of more special characters in front/behind of generics like `in` or `const` which are not valid in the `<|>` position I'm leaning more and more towards switching it towards `generics=".."` on the `<script>` tag.

wvhulle:
@dummdidumm, why are in and const not valid within the <script> tag? How are you going to fit the const and the in on the generics property? If you have more complicated types as generics, it seems like it would more reader-friendly to keep them next to the other parts of the interior of the typescript code inside the script tag. The way it has been until now, was great. If only the new eslint-parser could follow up.

AlbertMarashi:
I think putting generics in the `script` attributes makes sense, anywhere else is weird/odd since I see the `script` as a sort of function with generics.

Generics are always gonna be relatively simple in their definition... but @dummdidumm what if we needed to access a local type such as:

```svelte
<script generics="T extends LocalThing">
import LocalThing from "./LocalThing"
```
how would that work?

dummdidumm:
> Why are in and const not valid within the <script> tag? How are you going to fit the const and the in on the generics property? If you have more complicated types as generics, it seems like it would more reader-friendly to keep them next to the other parts of the interior of the typescript code inside the script tag. The way it has been until now, was great. If only the new eslint-parser could follow up.

@wvhulle [this is what I mean by const/in being invalid](https://www.typescriptlang.org/play?#code/PTAEBcHtQWwQwNYFNQEtygCZIGaoHZKYA0o+kGADnAE4aQ6gDGkM1NqAzpPgLABQ4AJ6UUAEjEBxJIQ5MAPAEFQAXlBx8QgHyrQigNwCBIZj04YAKgJwBXfE3CoeoHJEjyW+c6AtaAFOxwMABcPgCUoQBukKiYoADeAL6gJpAIRoIiKABibgAKtEG6EtKyqAqe3r76KWCcQvjgcAAeoEg0NJA0GSYEPtZ2Dk74oABGtPJ9vgGFIeFRMXFJtaBpGcKioABCtAU0RWolMu3lkyPVK-WNLW0dXT1gSM3gMpicGbb2js7jAF7yFjaz1enFA5g4+AA5v5wKELBFQNFYglkql0vwBBsUDtfnsDqAjmUFOCCNCVmlSKMbBgaEgAI42VC00G0gDuTNJoCEkBsNDG+z6UFACHIrIgAAs4BhwOKuLAkBpQfIcXiYECXvg3mDwBDoQIgA). It also shows that while yes it works with `$$Generic`, it's less readable for a newcomer already familiar with generics in TypeScript because it looks different. Even me, when I read `type T = $$Generic<string>` I always catch myself stopping for a moment to transform the syntax in my head to know that this is a generic that extends `string`. That wouldn't be the case for `<script generics="T extends string">..`

brunnerh:
Access to other types as mentioned by @AlbertMarashi is definitely an issue, personally I would rather *not* have the mess of:

```html
<script generics="T1 extends import('./path.ts').FirstType, T2 extends import('./path.ts').SecondType">
```

It should work for external types, but is really ugly, redundant/verbose. It also would *not* work for any local type definitions (as in a `type` or `interface` defined within the script), so all of those would have to be extracted to separate files which as a hard requirement could be a bit of an annoyance.

Further, any valid type has to be supported, such as utility types like e.g. `Omit<...>` and `Pick<...>`, but that probably is not an issue if the attribute is just extracted and turned into a TS fragment.

GauBen:
I use generics to have events typed using a children component's props, it looks roughly like this:

```svelte
<script lang="ts">
  type Entry = $$Generic<
    ComponentProps<EntryComponent> & {
      id: string;
    }
  >;
  
  export let entries: Entry[];
  
  const dispatch = createEventDispatcher<{
    click: Entry;
  }>();
</script>

{#each entries as entry}
  <EntryComponent {...entry} on:click={() => dispatch('click', entry)}  />
{/each}
```

I don't know how much of a pattern/anti-pattern this is, but it allows for complete type inference without hand-maintained types

dummdidumm:
The rule is: Everything that is imported or present in `<script context="module">..` is available. Example:
```svelte
<script context="module" lang="ts">
  interface Foo { bar: string } // available in generics because module context is available across components
</script>

<script lang="ts" generics="T extends Foo, X extends Bar">
  import type { Bar } from './somewhere'; // available in generics because imports are top level
  
  export let t: T;
  export let x: X;

  interface Baz { d: boolean } // not available
  // ...because you could also something like
  export let prop: string;
  interface WeirdButWorks { [prop]: typeof prop }
</script>
```

In the case of interfaces this might require you to type a few more characters because you need an extra context module, but it's a clear win in clarity. Right now, it's totally unclear what you can put inside `$$Generics<..>` or where `$$Generics` is hoisted to. With the `generics` attributes it's clear that the things the generics use need to be available outside of a component instance.

The case of `interface Baz { foo: string }` _could_ be hoisted out if we detect it does not depend on any local instance variable, so maybe we can even lift that limitation later on, though it might sacrifice a bit of clarity then. In practice it's probably rare that you define an interface only to extend a generic from it.

AlbertMarashi:
I'm happy w this approach then. Generics attribute makes most sense and is closest to how you would put generics on a function.

Since it can handle Importing types via hoisting/top-level import this solves most of my personal issues

TGlide:
I personally prefer the `$$Generic` syntax, if it has no compromise in functionality. It feels less constrained than having to use it inside what is basically an html attribute. Also wondering how the `generics` attribute would look like and format with long/multiple types

wvhulle:
If the generics version works well and can do all the same or more, it is fine for me. But I cannot just throw away my existing generic components (which have been used in production for almost a year) until this is resolved, so I reverted to eslint svelte3 which still works as expected. 

Apart from that, @dummdidumm sample code seems to have valid arguments for the generics attribute. 

I understand the $$Generic angular bracket syntax is not very consistent with other Typescript language constructs. Isn't there an alternative to using an html attribute for this? Perhaps an @ sign or something else that makes it clear this is not just a normal html property or JavaScript?

Maybe keep both options, $$Generic and generics open?

Nick-Mazuk:
How would records in the generic attribute work?

```svelte
<script lang='ts' generics='T extends { foo: string }'>

</script>
```

Normally curly braces in HTML mean to substitute the contents with with JS expression. But here we don't want the substitution and `foo: string` is not a valid JS expression.

FWIW, I tried this with the change in https://github.com/sveltejs/language-tools/pull/2020 released in v3.4.3 and am getting the following error:

```
Error: Expected } (svelte)
<script lang="ts" generics="T extends { key: string }">
```

I could hoist the `{ key: string }` into the module script, but that seems like a workaround that shouldn't exist.

```svelte
<script lang='ts' context='module'>
    type Foo = { foo: string }
</script>

<script lang='ts' generics='T extends Foo'>

</script>
```

I already filed https://github.com/sveltejs/language-tools/issues/2039 to keep track of this but posting here as well in case we want to discuss the desired solution.

TheOnlyTails:
I imagine that you would have to escape the curly braces as normal with `&123;` and `&125;`.

brunnerh:
It probably does not make sense to try and interpolate here, the value should be constant at compile time.

> I imagine that you would have to escape the curly braces as normal with &123; and &125;.

I would much rather just do this, if really necessary:
```svelte
<script lang='ts' generics={'T extends { foo: string }'}>
```

TGlide:
Also, how would the generics attribute work with events & slots? Would it also be an attribute?

If so, it really worries me legibility wise compared to what we have now:
Screenshot 2023-05-30 at 17 52 29

dummdidumm:
The generics defined on the generics attribute are available to the whole instance script content, including `$$Events` and `$$Slots`.

The record thing is a good catch, the Svelte parser thinks it's a mustache tag when it shouldn't. Mhm we may need to special-case this inside the parser, or add a change to the typescript preprocessor where it strips the generics attribute. The latter is now easily possible in Svelte 4, so that's probably the most promising route.

wvhulle:
@dummdidumm will it be possible to use the imported types inside the script instance tag inside the generics attribute? 

And will it be possible to use type aliases defined in the script instance tag inside the generics tag?

TGlide:
> The generics defined on the generics attribute are available to the whole instance script content, including `$$Events` and `$$Slots`.
> 
> The record thing is a good catch, the Svelte parser thinks it's a mustache tag when it shouldn't. Mhm we may need to special-case this inside the parser, or add a change to the typescript preprocessor where it strips the generics attribute. The latter is now easily possible in Svelte 4, so that's probably the most promising route.

I see, thanks for the answer! Though I think I expressed myself badly, I meant to ask if $$Events and $$Slots will also be replaced by an attribute?

AdrianGonz97:
Hey folks,

I just wanted to note that I _love_ the new `generics` attribute for one particular reason that I have yet to see anyone comment on (I've read from March to today). 

I'm not aware of _**any**_ way to have [Generic Parameter Defaults](https://www.typescriptlang.org/docs/handbook/2/generics.html#generic-parameter-defaults) with `$$Generic` (if there's a way, I'd greatly appreciate being enlighten 😄).

Since Svelte components are compiled down into classes, we _should_ have a way to specify default generic types like so:
```ts
class SomeSvelteComponent<T extends SomeType = SomeDefaultType> { ... }
```
But I don't think `$$Generic` can permit that.

With the `generics` attribute, we can simply specify the default like so:
```svelte
<script lang="ts" generics="T extends SomeType = SomeDefaultType">
```
and it works beautifully! 
(the only exception being that if you put a `typeof someFunction` for the default type, the syntax highlighter gets all wonky)

ptrxyz:
I also have trouble getting things to work with `generics="....."`. Some examples include:

```jsx
// case 1: parsing error, expected `}`
<script lang="ts" generics="T extends { id: string }, Y extends string">

// case 2: unexpected end of input"
<script lang="ts" generics="T extends Record<string, number>, Y extends string">
```

In both cases the type does not resolve and ends up as `any`

ptrxyz:
I found this: https://github.com/sveltejs/language-tools/issues/2039

Nick-Mazuk:
Anyone else having issues with svelte-check not type checking props when using the generic attribute?

I filed https://github.com/sveltejs/language-tools/issues/2107 just in case, but figured someone here may have figured out a way to get it working.

tzuleger:
Is there hope for JSDOC typing to see support for this in the future?

mythril:
I'm just an end-user but, generics as a string attribute seems very out of place to me.

I don't see why it would be preferable to do that vs a new tag that has some restrictions on it, and essentially hoists the component into a class definition.

for example a context="type" tag which can only contain type info, a component attribute to give the component a name in the context to work with, a truncated (no body?) class definition, and a token to represent the generated component code
```typescript
<script lang="ts" context="type" component="SomeSvelteComponent">
  class SomeSvelteComponent<T extends SomeType = SomeDefaultType> { ...$$Component }
</script>
<script lang="ts">
  // regular component definition here
  export let value: string = '';
</script>
```

I think this is sort of like the first option "ComponentDefinition" approach, but it seems to me that it could almost be a text-replacement macro, applied after the component is generated, then type checked. Though I don't really understand the depths of the maintenance burden discussion, nor have I ever even looked at the compiler and how it works, so sorry if this is just noise.

mythril:
Now that I think of it this might provide a path for extendable components.

mythril:
And alternate type systems like flow, jsdoc, and typescript

realfakenerd:
> I'm just an end-user but, generics as a string attribute seems very out of place to me.
> 
> I don't see why it would be preferable to do that vs a new tag that has some restrictions on it, and essentially hoists the component into a class definition.
> 
> for example a context="type" tag which can only contain type info, a component attribute to give the component a name in the context to work with, a truncated (no body?) class definition, and a token to represent the generated component code
> 
> ```ts
> <script lang="ts" context="type" component="SomeSvelteComponent">
>   class SomeSvelteComponent<T extends SomeType = SomeDefaultType> { ...$$Component }
> </script>
> <script lang="ts">
>   // regular component definition here
>   export let value: string = '';
> </script>
> ```
> 
> I think this is sort of like the first option "ComponentDefinition" approach, but it seems to me that it could almost be a text-replacement macro, applied after the component is generated, then type checked. Though I don't really understand the depths of the maintenance burden discussion, nor have I ever even looked at the compiler and how it works, so sorry if this is just noise.

I think that doing something like this would be better or making the types available inside the `"./$types"` import. Maybe whatever you write inside the `context="type"` is put inside the generated typescript file, so it could be imported by another components, i was looking for a way of getting inference through the context api.

It would be great if we used `setContext<SomeType>('somekey', something)` and it would automatically get the type when i do `getContext('somekey')`.

PR #39, opened 2020-09-27
Better slots and component composition.

TheComputerM:
[Rendered](https://github.com/TheComputerM/rfcs/blob/thecomputerm/slots/text/0000-better-slots.md)

This is my first RFC so please comment any helpful suggestions.

### DIFF
--- a/text/0000-better-slots.md
+++ b/text/0000-better-slots.md
@@ -1,0 +1,161 @@
+- Start Date: 2020-09-27
+- RFC PR: (leave this empty)
+- Svelte Issue: (leave this empty)
+
+# Better Slots and Component Compositon
+
+## Summary
+
+This RFC aims to provide a better way to create nested components and improve component composition using slots. This also aims to provide a more *svelte* like syntax for creating slots.
+
+**Input**
+```html
+<div class="has-slot-named">
+  <svelte:slot name="named">
+    Default Content
+  </svelte:slot>
+</div>
+<svelte:slot>
+  This is the default slot.
+</svelte:slot>
+```
+**Component**
+```html
+<svelte:fragment slot="named">
+  Some Content
+  <SomeOtherComponent />
+</svelte:fragment>
+```
+**Output**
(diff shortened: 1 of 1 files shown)


TheComputerM:
Somewhat like https://github.com/sveltejs/svelte/pull/4556

lukeed:
I was just going to link that – what are the differences? It looks like this is implemented as described

TheComputerM:
This is still incomplete and I am brain storming on better syntax.
The goal of it is to provide a more svelte like syntax and some additional features also such as slot options:
```html
<Component>
  <svelte:fragment dense={true}>
    This will do something in the component
  </svelte:fragment>
</Component>
```

lukeed:
Wouldn't slots just be a portal for _where_ to render? The contents of _what_ is rendered is determined by the content itself.

As far as I understand it, this RFC seems to be mostly about using `svelte:fragment` to render text nodes (non-component and non-element) with siblings.

I have an RFC open about inheriting `<slot/>` attributes (#15)
But (imo) if you're passing `dense={true}` so that the slot can react/alter its display, then I think that falls more into `svelte:template` territory (#34).

But, I might be misunderstanding you completely. I'm sorta guessing at the examples

TheComputerM:
Yeah, sorry for the short explanation. Was thinking more on the lines of something like this:
The syntax is just for better understanding.
```html
<svelte:slot tag={ListItem} export:dense on:click={func}>
  Default
</svelte:slot>
```

```html
<svelte:fragment dense={true}>
  Content
</svelte:fragment>
```

```html
<ListItem dense={true} on:click={func}>
  Content
</ListItem>
```

TheComputerM:
This was mainly intended to solve https://github.com/sveltejs/svelte/issues/2079 and add some new features.

lukeed:
No problem

Yeah, to me that's 100% an inline component, if not an external component. Either of which only get _placed_ by a `<slot>`

PR #41, opened 2020-11-25
Declarative Actions

Zizico2:
[Rendered](https://github.com/Zizico2/rfcs/blob/master/text/0000-declarative-actions.md)


aewing:
I think it might be nice to use slots within the target and consumer to allow specific manipulation of the DOM. 
What if the default behavior of an empty `<target />` was to contain a single `<slot />`, and the default behavior of a consumer of an action was to render to the default slot, but when one or more named slots appear in a target inside a template, targeting named slots within the consumer also works?
Not sure about the feasibility here, but I could imagine a scenario with `use:ripple` where it might be nice to append an `svg` to the consumer. I'm sure there are other more complicated scenarios where it might be nice to selectively manipulate the contents of the consumer.

aewing:
I'd written a comment about potentially using slots within targets and consumers... then I started wondering if the target was necessary at that point - if that was the case, the request might be summarized as "use components as actions".
It might not be a bad thing to have syntactical sugar to wrap a component with others. Could this perhaps be a simpler way to characterize this request, which would also enable manipulation of the DOM, and relieve the need for an additional `target` element?

aewing:
It almost seems like this could be solved by wrapping the element with a component and slot props -- maybe use:Component is just syntactical sugar for wrapping an element with a component?

Zizico2:
> It seems like this is already possible by wrapping the element with a component and slot props -- maybe use:Component is just syntactical sugar for wrapping an element with a component?

How would slot props solve the problem? And a component can't really assume its slot only has one child.

aewing:
> > It seems like this is already possible by wrapping the element with a component and slot props -- maybe use:Component is just syntactical sugar for wrapping an element with a component?
> 
> How would slot props solve the problem? And a component can't really assume its slot only has one child.

How would they not? What would this make possible that would not otherwise be possible by wrapping an element with a component? I mention slot props because `use:Component` might assist in facilitating the flow of data there.

To your point about a slot containing multiple child nodes, perhaps in the context of `use:Component` we could expect one or more nodes, or throw an error for multiple child nodes. Taking an array might not be a bad thing... what if I wanted the action to apply to all child nodes?

Zizico2:
> > > It seems like this is already possible by wrapping the element with a component and slot props -- maybe use:Component is just syntactical sugar for wrapping an element with a component?
> > 
> > 
> > How would slot props solve the problem? And a component can't really assume its slot only has one child.
> 
> How would they not? What would this make possible that would not otherwise be possible by wrapping an element with a component? I mention slot props because `use:Component` might assist in facilitating the flow of data there.
> 
> To your point about a slot containing multiple child nodes, perhaps in the context of `use:Component` we could expect one or more nodes, or throw an error for multiple child nodes. Taking an array might not be a bad thing... what if I wanted the action to apply to all child nodes?

Check out the "More on styling" section of the RFC. How would you do that with a wrapper component and slot props? I have an example with a wrapper Component. Applying styles to the parent is not the same as applying styles to the Element itself. How would you solve the problem I propose there? Could you spin up an example on the REPL?

aewing:
> > > > It seems like this is already possible by wrapping the element with a component and slot props -- maybe use:Component is just syntactical sugar for wrapping an element with a component?
> > > 
> > > 
> > > How would slot props solve the problem? And a component can't really assume its slot only has one child.
> > 
> > 
> > How would they not? What would this make possible that would not otherwise be possible by wrapping an element with a component? I mention slot props because `use:Component` might assist in facilitating the flow of data there.
> > To your point about a slot containing multiple child nodes, perhaps in the context of `use:Component` we could expect one or more nodes, or throw an error for multiple child nodes. Taking an array might not be a bad thing... what if I wanted the action to apply to all child nodes?
> 
> Check out the "More on styling" section of the RFC. How would you do that with a wrapper component and slot props? I have an example with a wrapper Component. Applying styles to the parent is not the same as applying styles to the Element itself. How would you solve the problem I propose there? Could you spin up an example on the REPL?

The use of `:global` isn't exactly awesome, but there are ways:
https://svelte.dev/repl/7b0c6c778aa04b2da0e862a71721c187?version=3.30.0

> Then it wouldn't be an action anymore.
True.

Zizico2:
> The use of :global isn't exactly awesome, but there are ways:
> https://svelte.dev/repl/7b0c6c778aa04b2da0e862a71721c187?version=3.30.0

Ah, I see. One of my intentions with this RFC was to avoid `:global` as with global css there are, indeed, ways.
That solution is in fact better looking than the ones I proposed as possible alternatives. I'll add the `PassiveModifier` (even thought it assumes a single child) approach to the the "More on Styling" section when I have some time, probably later today.
(there might be a way to style parent Elements in the future, [:has](https://drafts.csswg.org/selectors-4/#relational))

But apart from styling I still do think Declarative Actions are a generally better way to write Actions.

aewing:
I can certainly see the appeal, I was just thinking it might be possible to deliver the reactivity you seek without introducing the `<target />` element and a special paradigm for reactive actions by leveraging components instead.
It's too bad you can't `bind:this` to a component's `let:` declaration, or this reactivity would be a lot easier with a component wrapper. But what if you could? What if a component could assert that it only takes a single child? What if onMount (or some other mechanism) provided the refs of child elements? Might that be less complicated a change to deliver this ask?

Zizico2:
There might be other ways. Introducing a new Element (or using `<slot />`) will always be necessary for styling though, since Svelte will delete non used styles. Without relying on global styling that is.

dummdidumm:
One thing that is not clear to me is how well the proposed type of actions will be composable. A big strength of the current actions (and a reason to keep the "old" way in, even if this RFC would be implemented) is that they are just functions and can easily be mixed and matched with other functions or even actions.

isaacHagoel:
Hi, 
Sorry for being late to the party. I am the author of the most popular (afaik) drag and drop library, [svelte-dnd-action](https://github.com/isaacHagoel/svelte-dnd-action), which uses an action (that is actually aggregating two other actions).
The team that I am leading at work also uses svelte for some of our production apps and we write custom actions very frequently. 
imo, actions are one of the most powerful and unique features svelte offers (and in many ways svelte's answer to React hooks) and replacing it would be a huge step backwards. The define a very clean interface for how javascript code can interact and be integrated seamlessly with svelte applications. The author of the action gets (almost) unlimited control and the consumers of the action get a tiny, elegant API.
Here is when we reach out for actions in their current form (outside of the obvious usecases):
1. actions in their current form are not limited to reasoning about a single component. they can be used to tie components together (ex: holding a map with references to each component that uses the action and triggering events on any of them whenever it sees fit). It can address infinitely complex usecases yet, from the perspective of the consumer of the action, there is a nice, compact, declarative API that completely encapsulates all of the implementation details. 
2. When we want to implement functionality that can be easily ported to be used with any library/ framework/ vanilla js. Because actions don't need svelte as a dependency they can easily be wrapped with a React hook or be used directly (calling update when needed) in any codebase.  I experimented with it [here](https://github.com/isaacHagoel/react-dnd-action) and we do it for one of the internal libraries we've created at work.
 
If you go with this please do it in addition and not instead the existing API.

lukaszpolowczyk:
@IsaacHagoel I would never have thought that this proposal is to replace the existing API.
But I'm not a decision maker, just a fan and user of Svelte.

Zizico2:
> One thing that is not clear to me is how well the proposed type of actions will be composable. A big strength of the current actions (and a reason to keep the "old" way in, even if this RFC would be implemented) is that they are just functions and can easily be mixed and matched with other functions or even actions.

Composition should be fairly simple since since `<target />`, or any other Element for that matter, can have multiple actions applied to it.

```html
// blueText.svelte
<style>
.blue {
    color: blue;
}
<style class:blue />
<target />
```

```html
// redBg.svelte
<style>
.red {
    background_color: red;
}
<style />
<target class:red />
```

```html
// composition.svelte
<script>
import { redBg } from "redBg.svelte";
import { blueText } from "blueText.svelte";
</script>
<target use:redBg use:blueText >
```

As for interoperability with any regular function. I proposed having lifecycle hooks that allow us to write purely imperative actions like before if necessary. 

```html
// regularOldAction.svelte
<script>
import { init, destroy } from "svelte/actions";

let node;

init(() => {
    console.log("Hi!")
    node.innerHTML = "An old action is being used!"
});
destroy(() => {
    console.log("Bye!")
});
</script>
<target bind:this={node} />
```

This would also address some of @isaacHagoel 's concerns. If need be, you can write actions like before. My idea was, indeed, to replace the current actions but the lifecycle hooks would provide an option for purely imperative actions. The `mount` and `destroy` hooks would be fairly easy to port over from current actions. The current `update` function would need to be rethought to use reactive declarations and props. I understand this is more convoluted than the current solution. It would, indeed, be a tradeoff.

As for this:

> When we want to implement functionality that can be easily ported to be used with any library/ framework/ vanilla js. Because actions don't need svelte as a dependency they can easily be wrapped with a React hook or be used directly (calling update when needed) in any codebase. I experimented with it here and we do it for one of the internal libraries we've created at work.

I don't have a solution for this issue. An overall would be required. Assuming your use case wouldn't benefit at all from Declarative Actions (I have no idea if it would, pure assumption), and you would still write everything imperatively using the lifecycle hooks, you would need to abstract the functionality away, to a framework agnostic set of functions. Then you could wrap those functions with a react hook, a svelte action, whatever. Instead of: (React Hook (Svelte Action)) you would need to have (React Hook, Svelte Action (FW Agnostic Lib))
This increases Svelte's API surface, which I understand is not desirable.

Keeping old actions around could be an option but deciding between Declarative Actions and Imperative Actions would be a bit of a gray area.

lukaszpolowczyk:
What do you think?
Direct use of the `target` element:
```svelte
<!-- component.svelte -->
<script>
  let target;

  let isHeld = false;
</script>

<style>
  .red-background {
    background-color: red;
  }
</style>
<target
  class:red-background="{isHeld}"
  on:pointerdown="{() => {isHeld = true}}"
  on:pointerup="{() => {isHeld = false}}"
  bind:this="{target}"
/>

<div class="blue-text" use:target>
  I am declaratively styled by an action! yey
</div>
```

It would work like shortening the entry inside a component, something like the equivalent of this:
```svelte
<script>
	let name = 'world';
	let x = ""
	let ob = {
		style:"color:red;",
		"on:click": ()=>{x="click"}
	}
</script>

<h1 {...ob}>Hello {name}!</h1>
{x}
```
But `ob` doesn't work and the `target` element would work.

Of course, you could create multiple targets in one file.

This could be `<svelte:target>`

Zizico2:
@lukaszpolowczyk I don't quite understand what you mean. Would you mind giving another example? How would multiple targets work? When you apply an action to an Element, what would the multiple targets refer to? Actions would still be applied as they are now. Do you want to use target in Components as a "1 child slot"?

lukaszpolowczyk:
@Zizico2 "How would multiple targets work?"
Separate variables - `target` and `target2`:
```svelte
<!-- component.svelte -->
<script>
  let target;
  let target2;

  let isHeld = false;
</script>

<style>
  .red-background {
    background-color: red;
  }
</style>
<target
  class:red-background="{isHeld}"
  on:pointerdown="{() => {isHeld = true}}"
  on:pointerup="{() => {isHeld = false}}"
  bind:this="{target}"
/>
<target
  class:blue-background
  on:click={/* function */}
  bind:this="{target2}"
/>

<div class="blue-text" use:target>
  I am declaratively styled by an action! yey
</div>
<button use:target2>Button1</button>
<button use:target2>Button2</button>
```

It works like if the code was like this:
```svelte
<div class="blue-text" 
  class:red-background="{isHeld}"
  on:pointerdown="{() => {isHeld = true}}"
  on:pointerup="{() => {isHeld = false}}">
  I am declaratively styled by an action! yey
</div>
<button 
  class:blue-background
  on:click={/* function */}>Button1</button>
<button 
  class:blue-background
  on:click={/* function */}>Button2</button>
```
Basically, the idea is to collect multiple svelte attributes (`on:event` `class:name` etc.) for an item into one.
Because normal HTML attributes can be done without it - as I showed it from `{...ob}`, in the previous comment - with the normal HTML attributes do not have this problem. 

The `target` and `target2` actions can be exported, so this could be used in other components.
This could be a `*.svelte` file with only actions.

Then `<script context="action">` would not create the action, only the `<target>` element itself would create the action, combining all the attributes (functions, variables) and styles.

---

Now I thought that the problem with `init` `destroy` could be solved like this:

```svelte
<script>
let target;
target.init(() => {
    console.log("Hi!")
    node.innerHTML = "An old action is being used!"
});
target.destroy(() => {
    console.log("Bye!")
});
let target2;
target2.init(() => {
    console.log("Hi!")
    node.innerHTML = "An old action is being used!"
});
target2.destroy(() => {
    console.log("Bye!")
});
</script>
```

Just the `<target>` element would have its own life cycle functions, available via `bind:this` attribute. 

---

BTW: Another name could be `<svelte:action>`.

Zizico2:
Ah I see what you mean now. I'll give you my thoughts once I have time to settle down in front of a computer and I might update the RFC with this as an alternative. I quite like it if I'm being honest. But how would you implement arguments? And the current "update" function in actions? Just adding a "lifecycle:update" doesn't sound good as it defeats the point of "Declarative Actions". And using "bind:this" might be a bit misleading, that's what threw me off the first time I read it. "bind:action" or so would be more appropriate. But then how do you get access to the Element the action is applied to? An argument could be passed to the "init" and "destroy" functions. But what if you need to use it outside of those functions? The current way I solved arguments and "update" was to use props and reactive declarations. That's as declarative as it gets and it borrows the system from Components so it would be intuitive to teach.

lukaszpolowczyk:
@Zizico2 "t. But how would you implement arguments?
"I've thought over my idea, and it doesn't make sense.
Parameters cannot be used in a meaningful way.

Only your idea makes sense when it comes to a fully declarative solution.

Sorry for the confusion.

Maybe you can combine your idea with the Inline Components idea: https://github.com/sveltejs/rfcs/pull/34

It would look like this:
```svelte
<svelte:template name="image" context="action" let:image>
  {@const isHeld = false;}
<target
  class:red-background="{isHeld}"
  on:pointerdown="{() => {isHeld = true}}"
  on:pointerup="{() => {isHeld = false}}"
  bind:this="{target}"
/>
</svelte:template>

<div class="blue-text" use:image={image}>
  I am declaratively styled by an action! yey
</div>
```

What to `use:image` is the problem. With Inline Components you use `<svelte:component this="image" {image}>`  and there is no equivalent for `use:action`. 

You would have to change `name="image"` to e.g. `<svelte:template component={Image} let:image>`, where Image is an object-variable. Then it is easy to use `use:Image`.
But I don't know if this is a good idea, because it uses `<svelte:component this="image "{image}>` for a reason.

---

A slightly different idea:
You could just extend the standard `use:action`:
```svelte
<script>
let action;
action.init(node, parameters)=>{
return {update:(newParameters)=>{},destroy:()=>{}}
}
<script>
<target
  class:blue-background
  on:click={/* function */}
  bind:action="{target}"
/>
<button use:target>Button1</button>
<button use:target>Button2</button>
```
But this is not a fully declarative solution.
This is something in between - it is not fully declarative, but it has e.g. `on:click` `class:name` etc. and there can be many actions in one file (just like the standard `use:action`).

But I'm not attached to this idea.
Probably combining Declarative Actions with Inline Components would be better.

Zachiah:
This would be amazing. A major use case would be when you're using svelte with web components, because styles don't flow down in web components using these type of actions for styling would be very helpful.

lukaszpolowczyk:
@Zizico2 I had a crazy idea ...
I had a crazy idea ...

That would be counter to JS's intuition and logic, but would solve the problem.
And with svelte logic it would be perhaps even quite intuitive.

On the example of this changed:
https://github.com/sveltejs/rfcs/pull/41#issuecomment-798795216

The isHeld variable would be created multiple times with the "salted" (like CSS classes) property name, on subsequent `use:target2` writes - example of behavior: https://svelte.dev/repl/0e811865dacd4ba3a6ca3fe419e9f455?version=3.38.2
And it would be reactive with the original variable, so the effect would be:

```svelte
<!-- component.svelte -->
<script>
  let target2;

  let isHeld = false;
  $:isHeldS54JK4DF67IM = isHeld; // Generated
  $:isHeldSH5DF45THEX = isHeld; // Generated
</script>

<style>
  .red-background {
    background-color: red;
  }
</style>

<!--  Generated something like this duplicate something, but it is known that in the compiled Svelte code it would be something else...
<target
  class:red-background="{isHeldS54JK4DF67IM}"
  on:pointerdown="{() => {isHeldS54JK4DF67IM = true}}"
  on:pointerup="{() => {isHeldS54JK4DF67IM = false}}"
  bind:this="{target}"
/>
<target
  class:red-background="{isHeldSH5DF45THEX}"
  on:pointerdown="{() => {isHeldSH5DF45THEX = true}}"
  on:pointerup="{() => {isHeldSH5DF45THEX = false}}"
  bind:this="{target}"
/>
 -->

<div class="blue-text" use:target>
  I am declaratively styled by an action! yey
</div>
<div class="blue-text" use:target>
  I am declaratively styled by an action! yey
</div>
```

So the variables used inside the `<target>` tag would be replaced with salted variables inside it.

You can then set isHeld from outside or from the overall body of the component. And by changing the variable from the `<target>` tag, only the salted variable would change.

**What do you think?**

I am thinking of unusual cases that could get in the way.
The functions would not have to be salted.
Perhaps there is something I have overlooked.

Prinzhorn:
First of all a big thumbs up for declarative actions. When I first started using Svelte and wrote my first action I thought "amazing, this is my interface into the regular DOM world and I can do things that are otherwise not possible or much harder with the regular component interface". After writing a couple more actions I realized that I'm feeling the same pain I felt with vanilla js / jQuery: keeping things in sync myself using imperative code. I found myself doing this:

```js
function myAction(node, options) {
  // ... actual code ...

  function update(newOptions) {
    // Is this cheating?
    // Svelte reuses the same node for different components, we basically just act like a new action was initialized.
    destroy();
    myAction(node, newOptions);
  }

  function destroy() {
    // ... undo things ...
  }

  return {
    update,
    destroy,
  }
}
```

Because figuring out what to do based on the diff between `options` and `newOptions` is exactly what I don't want to do manually and it's the job of a computer. I'm tired of simulating a computer in my head (thanks Svelte for doing that for me for the most part).

I think in three years we'll look back and think "why did we ever write imperative actions?"

There are a couple of smaller changes I'd make to this rfc, e.g. `<svelte:target>` instead of `<target>` and I didn't spend time to read everything it detail. But here's main gist:

**What keeps us from compiling such an action-component into the exact same format that actions currently use (with `'svelte/internal'` imports)?** All it has to do is use the same invalidation logic it currently does inside `update` and magically all operations that use `options` are invalidated (current `options` argument is literally `export const options`). All other things (event listeners, styles) can be compiled to the exact same primitives that already exist, e.g. `set_style(node)`. And the logic for cleaning up (e.g. unbinding events, removing classes) already exists in svelte, just put it in `destroy`. 100% backwards compatible with the current actions API, but the Svelte way.

Zizico2:
@Prinzhorn I'm not the right person to comment about your last paragraph, I'm really not that well versed in Svelte's internals.

And I will put `<svelte:target>` as an option in the rfc. I chose for `<target>` because `<slot>` isn't `<svelte:slot>` either, even though it doesn't behave as vanilla's JS/HTML `<slot>`. But I guess the tag does exist so they just used it.

dummdidumm:
One open question for me is how to make this composable. Right now you can mix and match actions to your liking

```javascript
function actionA(..) {..}
function actionB(..) {..}

function actionC(..) {
  actionA(..);
  if (something) {
    actionB(..);
  }
}
```

Another open question is how to make a "higher order" action?

```javascript
function createDragNDrop(..) {
  function drag(node) {..} // applied to the drag target
  function drop(node) {..} // applied to the drop target
  // logic for managing drag and drop zone in here, which can be used by both functions
  return { drag, drop }
}
```

How would these use cases translate to declarative actions?

If you want the goodies of Svelte components, what's stopping us from creating a Svelte component inside an action?

`action.js`
```javascript
import Action from './Action.svelte'

export function action(node, {text}) {
  const component = new Action({target:node, props: {node,text}});
  return {
    update: ({text}) => component.$set({text}),
    destroy: () => component.destroy()
  }
}
```

`Action.svelte`
```svelte
<script>
  export let text;
  export let node;
	
  node.style.backgroundColor = 'pink';
</script>

{text}
```

lukaszpolowczyk:
> creating a Svelte component inside an action

And then how do you add `on:eventname`, `class:name`, `transition:fn` etc. from the action/`Action.svelte` level?

Prinzhorn:
> One open question for me is how to make this composable. Right now you can mix and match actions to your liking

If declarative actions are compiled to the same function schema they use now, you could do one of these two:

1. Have your "meta" action _not_ be a declarative action (but a "classic" imperative one) and import the two declarative actions and use them as usual (once imported they're just a function).

```javascript
import actionA from 'actionA.svelte';
import actionB from 'actionB.svelte';

function actionC(..) {
  actionA(..);
  if (something) {
    actionB(..);
  }
}
```

2. Allow `use` on `<svelte:target>` so you can go full circle and bath in declarativity. So one declarative action can just add more actions using `use:`. If you want to do that conditionally we end up here https://github.com/sveltejs/svelte/issues/6754

> Another open question is how to make a "higher order" action?

See 1. above. Nothing keeps you from re-exporting the declarative action and bundle it with an imperative action.

> If you want **the goodies** of Svelte components, what's stopping us from creating a Svelte component inside an action?

\*\*uses `node.style.backgroundColor` to demonstrate goodies\*\* sorry :smile:

Zizico2:
How would you suggest we keep both types of actions? A declarative action would be usable with `use` or as regular ol' function? Enabling `use` on `<svelte:target>` is a really clean and succinct solution. I had already answered the composition concerns in an earlier comment with this exact solution. @dummdidumm

> Composition should be fairly simple since since <target />, or any other Element for that matter, can have multiple actions applied to it.
> 
> 
> <style>
> .blue {
>     color: blue;
> }
> </style>
> <target class:blue />
> 
> 
> <style>
> .red {
>     background-color: red;
> }
> </style>
> <target class:red />
> 
> 
> <script>
> import { redBg } from "redBg.svelte";
> import { blueText } from "blueText.svelte";
> </script>
> <target use:redBg use:blueText />

Prinzhorn:
> A declarative action would be usable with `use` or as regular ol' function?

Yes. If you currently `import Foo from Foo.svelte` you get a constructor and can do either `<Foo>` or manually `new Foo()`. If you import a declarative action (a component that has a `context="action"` block) doing `import foo from foo.svelte` you will get a function (not a constructor) that is indistinguishable from any regular action you wrote imperatively by hand. The module just happens to import from `svelte/internal` if needed. (So I'm pretty sure you could compile a declarative action to vanilla JS and use it in React if you really wanted to just like you can with a Svelte component. I don't.)

For me one open question is if this should have a different extension (e.g. `.action`). But these are so similar to regular components that it would be weird and tooling like the prettier plugin could already format this right now. **The question would be how we teach this** if the only different between declarative component and regular component is the existence of `context="action"`? For the compiler it's not an issue, if there is `context="action"` it will allow `<svelte:target>` and fail if it sees a regular `<script>` without `context` (like I said above `context="module"` would work for actions identical to component right now by scoping it outside the function).

Zizico2:
> Edit: The big issue with DX/teaching is that from the import statement you cannot distinguish between action and component, but both are used very differently. However, now that I think about it, that's true for literally every .js file you import. It's your responsibility to actually import a function that corresponds to the action spec, the compiler doesn't know.

This could just come down to a naming convention. `UpperCamelCase.svelte` for components and `lowerCamelCase.svelte` for actions.

Prinzhorn:
> What about actions without a script at all?

Good point, maybe the existence of `<svelte:target>` triggers it, because that's what an action requires. It's the point of an action.

I realized that declarative actions would change my mental model for actions.;

* Currently: actions are my escape hatch into the DOM world and I can do with the node whatever I want
* New: actions allow _horizontal extension_ of elements. In contrast to the component system (with composition/context etc.) which are all _vertical_.

Once I had this new mental model three things came to my mind:

1. "Should we allow more Elements than the <target /> Element in an action?" gets a very clear **no** from me. That's not what actions are for (that would be a vertical change) and once you allow that it becomes even harder to distinguish components from actions. And it would bring a lot of new problems when a component plus multiple actions want to change the tree (who would win?). With attributes it's just a matter of having a well defined order (e.g. setting an attribute in an action overrides the one in the component). 

2. Declarative actions would solve another problem I currently have: utility first CSS frameworks. I'm using Tailwind for the first time and it doesn't pair well with Svelte. I don't want to extract things into components just for the styling, for me components need to _do_ something. But even if I extract them things get worse. E.g. I cannot use `animate` because now the element (styled component) is not an immediate child of the keyed each. With declarative actions I could just style my element using actions (with the current actions that would just result in tons of ugli `toggleClass` calls). It would also keep things DRY, because currently my styled `<Button>` has a path for `<a>` and one for `<button>` with basically identical styles.

3. Maybe we can teach it using the vertical/horizontal terminology?

Regarding the parameters: I think we stick with the current approach, because declarative actions have the current action spec as compile target (they need to be compiled to _something_ anyway). So the parameters become `export const parameters` in the declarative component. Doesn't matter what you put in it (a primitive value, function or object) it is reactive in the same way props are right now. If you pass an object and use `parameters.foo` somewhere Svelte will do it's surgical updates so that you don't need to do it (that's my main motivation for declarative actions, I want consistency and not diff `parameters` and `newParameters` in `update` manually).

If the maintainers could signal if this is something they can see happening I'll go ahead an review the whole rfc. I still haven't read it in it's entirety.

Zizico2:
We shall wait.

And I don't think I understood this: 
> Regarding the parameters: I think we stick with the current approach, because declarative actions have the current action spec as compile target (they need to be compiled to something anyway). So the parameters become export const parameters in the declarative component. Doesn't matter what you put in it (a primitive value, function or object) it is reactive in the same way props are right now. If you pass an object and use parameters.foo somewhere Svelte will do it's surgical updates so that you don't need to do it (that's my main motivation for declarative actions, I want consistency and not diff parameters and newParameters in update manually).

What's stopping us from allowing multiple props, like in a Component, and then compiling them into a single function parameter? This could be less intuitive then just forcing actions to have a single prop and letting the user handle that themselves I guess, but it would sure be more ergonomic.

Prinzhorn:
> What's stopping us from allowing multiple props, like in a Component, and then compiling them into a single function parameter? This could be less intuitive then just forcing actions to have a single prop and letting the user handle that themselves I guess, but it would sure be more ergonomic.

So what you're suggesting is that actions compiled from declarative action _always_ require the argument to be an object? Because right now I have actions that get a single function as argument.

You could always do this if you want to:

```svelte
<script>
export let parameters;

$: {foo, bar} = parameters;
</script>
```

This would be 100% backwards compatible. If you you want this:

```svelte
<script>
export let foo;
export let bar;
</script>
```

Then you need to change actions that currently don't use an object if you want to migrate to declarative actions.

But there could also be a magic `$$param` or something that would hold my function and in this case I wouldn't declare any exports?

lukaszpolowczyk:
Such a loose thought:

Isn't Declarative Actions more of a **component + [svelte:element](https://github.com/sveltejs/svelte/issues/2324) + additional goodies**?

1. declarative action:
```svelte
<!-- action.svelte -->
<script>
</script>
<target on:click={_=>{}} />
```

```svelte
<!-- App.svelte -->
<script>
   import { action } from './action.svelte';
</script>
<div use:action {...individualPropsdiv>
```

vs

2. svelte:element:
```svelte
<!-- action.svelte -->
<script>
   export div;
</script>
<svelte:element tag="div" on:click={_=>{}} {...$$restProps}><slot/></svelte:element>
```

```svelte
<!-- App.svelte -->
<script>
   import { action } from './action.svelte';
</script>
<action tag="div" {...individualProps}></action>
```

What difference?
In the First, you can use `on:click` etc. in both `action` and `App`.
In Second, you can use `on:click` etc, only in `action` (or get a bit more tiring with `createEventDispatcher`, for example).

So actually, you need to set `on:click`, `bind:val`, `class:name` etc. dynamically?
Maybe some `$$specialProps` and support for it in `svelte:element`? Then it would work:

3. svelte:element +`$$specialProps`:
```svelte
<!-- action.svelte -->
<script>
   export div;
</script>
<svelte:element tag="div" on:click={_=>{}} {...$$restProps} {...$$specialProps}><slot/></svelte:element>
```
```svelte
<!-- App.svelte -->
<script>
   import { action } from './action.svelte';
</script>
<action tag="div" {...individualProps} on:click={_=>} class:name bind:val></action>
```

Attributes `on:click={_=>}` `class:name` `bind:val` would be hidden in `$$specialProps`? It would not be an ordinary object, but a special object that retains svelte attributes.
The component itself would have to allow `on:click` `class:name` etc. - or you would have to enable support via e.g. `<Component mode="action"/>`?

Summary:
I don't know if that would solve anything, maybe it would be more universal?
Let this comment be treated, perhaps, as an inspiration for reflection.

Zizico2:
> So what you're suggesting is that actions compiled from declarative action _always_ require the argument to be an object? Because right now I have actions that get a single function as argument.

Ah... I was completely disregarding actions with a single parameter. I guess we could force having an object as a parameter. Maybe leaving it as it is, is better though. I would personally prefer the "forcing object as parameter" option. This may not be a popular opinion though.

iacore:
It seems like this syntax implicates dynamic injection of **any directive** at runtime. I feel like this will be both confusing and abused.

lukaszpolowczyk:
> It seems like this syntax implicates dynamic injection of any directive at runtime. I feel like this will be both confusing and abused.
What is this absurd statement?

This is super explicit and convenient and complements the features of Svelte.

The only problem, is that it doesn't require big changes in the compiler and doesn't heavily enlarge the resulting component. I guess it has to work so that the thing the compiler does, the runtime will do.
Do I understand that correctly?

lukaszpolowczyk:
### **UPDATE 2** 

A separate RFCS for this - https://github.com/sveltejs/rfcs/pull/68
"Targeted Slots - passing and mixing attributes and styles to component, access the slotted element from inside the component"

---

An alternative syntax that gives more possibilities, using slots, but also less graceful:

```svelte
<!-- App.svelte -->
<script>
import Com from "./Com.svelte";
</script>
<Com>
  <svelte:element slot="button" class="class1"></svelte:element>
</Com>
<style>
  .class1 {}
</style>
```
```svelte
<!-- Com.svelte -->
<script>
let target;
</script>
<slot name="button" targeted this="button" bind:this={target} class="class2" on:click={_=>{}}/>
<style>
  .class2 {}
</style>
```

Description:
* the `targeted` attribute in the `<slot>` tag causes the slot to accept html attributes, svelte special attributes, and can use `bind:this={target}`.
* using `<svelte:element` shows that it is possible to set `tagName` inside a component
* by using named slots, multiple targets can be used, not just one as in the original Declarative Actions proposal
* unfortunately, by using slots, the syntax is less graceful

I would very much appreciate your feedback, do you like it?

### UPDATE 1
I noticed one mistake.
Slots pass parameters by name, which causes two problems:
* `<slot targeted>` is treated as passing a boolean, which would be read in `<svelte:element let:targeted>`. We don't want that, because it would break the old syntax.
* Other attributes in `<slot>` are also passed, and read by `let:val`. This conflicts with the proposed treatment of these attributes, as attributes of an element.

Solution
* We need to replace the `targeted` attribute with, for example, `targeted:name`. Then there is no collision problem. From then on, attributes should be treated as attributes of the element. It also frees up the `name` attribute, which also someone may want to use in a different way than naming a slot....
* Passing the value to `let:val` would be done via this `targeted:name={ {val: "string"} }` and read `<svelte:element let:val>` - This may be a less pretty way, but I think it's worth it, given the benefits.

The same example in a new way:
```svelte
<!-- App.svelte -->
<script>
import Com from "./Com.svelte";
</script>
<Com>
  <svelte:element slot="button" class="class1"></svelte:element>
</Com>
<style>
  .class1 {}
</style>
```
```svelte
<!-- Com.svelte -->
<script>
let target;
</script>
<slot targeted:button  this="button" bind:this={target} class="class2" on:click={_=>{}}/>
<style>
  .class2 {}
</style>
```

Where `button` in `target:button` is the name of the slot.

And you can something like this:
```svelte
  <svelte:element slot="button" class="class1" let:val>{val}</svelte:element>
  ...
  <slot targeted:button={{ val: "string" }} 
    this="button" bind:this={target} class="class2" on:click={_=>{}} name="name1"
  />
```
Where the `name` attribute is in this sense - https://www.w3schools.com/tags/att_name.asp, not in the sense of a slot name.
And in the value of `targeted:button` you pass val to `let:val` in `<svelte:element let:val>`.

This is a compromise to release the attributes of the slot, and be able to use them as attributes of a dynamic element.

**DISCLAIMER:** If one believes that "slots should be left alone", one could call it different. But it would literally work as upgraded slots.

lukaszpolowczyk:
A minor note - Using Declarative Actions is a bit like using the `is` attribute on a Custom element declared in Web Components:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/is

This `<p use:wordCount></p>`, is equivalent to `<p is="word-count"></a>`. :D

I don't know, maybe this will encourage someone to pursue Declarative Actions. :)

obraz

madeleineostoja:
I don’t know what the is syntax has to do with this proposal, the is=“el” was also abandoned from the web components spec because safari wouldn’t implement it

lukaszpolowczyk:
@madeleineostoja Topic appears to be unfinished - https://bugs.webkit.org/show_bug.cgi?id=182671#c18


As for the lack of similarity - maybe I have more imagination.... I don't want the topic to be closed by a drawn out side discussion.

**In a nutshell:**
You have a database of attributes and parameters from an element + you add your own.
This is such an extension of the element behavior, INstead of wrapping the element.

Both the Declarative Action and the `is` attribute from Custom Element:
* increases code clarity
* allows you to extend the element in a simple way (Declarative Action uses the advantages of Svelte syntax, you do many things more simply than with use:action function)
* doesn't hide the properties and attributes of the component (and wrapping hides the properties of the component, and you have to extract them additionally to the component properties)

A regular `use:action`, where action is a simple function, also works like this, but the analogy is less because it doesn't use the advantages of the Svelte Component syntax, which means that extending the function is more difficult. (Element `extended by->` action function)

**Extension**, use:DeclarativeAction vs the `is` attribute:
Element `extended by->` Svelte Component
vs
Element `extended by->` Custom Element

**Wrapper**, putting Element in Svelte Component slot(or inside Component) vs putting Element in slot(or inside Custom Element):
Element ` wrapped by` Svelte Component
vs
Element ` wrapped by` Custom Element

**Summary:**
Declarative Action, is to use Svelte Component, without wrapping.
Attribute `is`, is to use Custom Element, without wrapping.

madeleineostoja:
Not going to get wrapped up in this, just wanted to mention that I was quite involved in spec discussions during the early days of web components, and I can assure you that customised built ins (`is=""`) will never see the light of day in Safari, which is why it was dropped from the final spec.

I don't understand what you're trying to get across, and there's no need for comments like "maybe I have more imagination"

lukaszpolowczyk:
> I don't understand what you're trying to get across, and there's no need for comments like "maybe I have more imagination"

I meant that maybe I associate things that don't fit together as I think they do.
It wasn't that you had too little imagination, I just had maybe "too much".

All in all, I feel bad that almost anything I write is received negatively.
I thought the analogy to something existing (in Firefox and Chrome), would help get people interested in Declarative Actions. I would love to see Declarative Actions implemented.

Zizico2:
> An alternative syntax that gives more possibilities, using slots, but also less graceful:
> 
> ```svelte
> 
> <script>
> import Com from "./Com.svelte";
> </script>
> <Com>
>   <svelte:element slot="button" class="class1"></svelte:element>
> </Com>
> <style>
>   .class1 {}
> </style>
> ```
> ```svelte
> 
> <script>
> let target;
> </script>
> <slot name="button" targeted this="button" bind:this={target} class="class2" on:click={_=>{}}/>
> <style>
>   .class2 {}
> </style>
> ```
> 
> Description:
> * the `targeted` attribute in the `<slot>` tag causes the slot to accept html attributes, svelte special attributes, and can use `bind:this={target}`.
> * using `<svelte:element` shows that it is possible to set `tagName` inside a component
> * by using named slots, multiple targets can be used, not just one as in the original Declarative Actions proposal
> * unfortunately, by using slots, the syntax is less graceful
> 
> I would very much appreciate your feedback, do you like it?

This is great. If I'm not forgetting anything this would be fantastic. I think I've alluded to a solution like this in a comment to an issue that I'm not finding right now xd. This might even be easier to teach since you could drop the notion of "Declarative Actions" completely. It's just a `targetted` slot in a "normal" Component.

Zizico2:
I've been really busy lately. If you want to you'd be welcome to open a pull request for my RFC, introducing this as an alternative. I'd happily review it. @lukaszpolowczyk

lukaszpolowczyk:
@Zizico2 I made a small but important UPDATE - https://github.com/sveltejs/rfcs/pull/41#issuecomment-1126434719
I detected a small problem with my syntax and suggested how to solve it.

And see this topic - https://github.com/sveltejs/rfcs/pull/60#issuecomment-1204651747
The idea is to solve Declarative Actions and Forward Directive at one time, through Targeted Slots.
What do you think?

lukaszpolowczyk:
@Zizico2 Check out my RFCS:

The syntax may not be as nice as `<div use:Component>`, but that makes it more comprehensive and standards-compliant (I hope).

mimbrown:
@Prinzhorn redirected me here after I posted a [similar issue](https://github.com/sveltejs/svelte/issues/7851). A couple thoughts, as I'm getting caught up and have been pondering this more:

* Just going to toss in `<svelte:composable>` as another alternative name.
* While @lukaszpolowczyk's RFC is quite interesting and powerful, it's not nearly as ergonomic as I would hope for this feature.
* While I think there should NOT be any other elements outside of the composable in its definition, I can actually see the ability to define children in a composable as both feasible and useful. Use case: material-style ripple (on click, add an element, animate the element, then remove).
* Actions, as they currently stand, may not be the right fit for compiled composability. The main reason is that they are not considered in SSR (see [this issue](https://github.com/sveltejs/svelte/issues/4375)). So if I defined a composable whose only job was to add a class (trivial example), then that class would not show up until hydration (not ideal). So maybe this is a completely new thing, not attached via `use:`. Possible alternatives (brainstorming):
```svelte
<!-- Given a composable named `dropZone` -->

<div with:dropZone />
<div compose:dropZone />
<div also:dropZone />
```

Zizico2:
> @Prinzhorn redirected me here after I posted a [similar issue](https://github.com/sveltejs/svelte/issues/7851). A couple thoughts, as I'm getting caught up and have been pondering this more:
> 
>     * Just going to toss in `<svelte:composable>` as another alternative name.
> 
>     * While @lukaszpolowczyk's RFC is quite interesting and powerful, it's not nearly as ergonomic as I would hope for this feature.
> 
>     * While I think there should NOT be any other elements outside of the composable in its definition, I can actually see the ability to define children in a composable as both feasible and useful. Use case: material-style ripple (on click, add an element, animate the element, then remove).
> 
>     * Actions, as they currently stand, may not be the right fit for compiled composability. The main reason is that they are not considered in SSR (see [this issue](https://github.com/sveltejs/svelte/issues/4375)). So if I defined a composable whose only job was to add a class (trivial example), then that class would not show up until hydration (not ideal). So maybe this is a completely new thing, not attached via `use:`. Possible alternatives (brainstorming):
> 
> 
> ```svelte
> 
> 
> <div with:dropZone />
> <div compose:dropZone />
> <div also:dropZone />
> ```

It's funny you mention the material ripple. That was the first time I felt the need for something like Declarative Actions adn then I generalized from there. And I don't think you would need to add an element to implement the material ripple. I think adding elements would make this RFC more convoluted than it needs to be and we would lose the differentiation between vertical and horizontal that @Prinzhorn conjectured in an earlier comment. You're right that using the current action infrastructure maybe wouldn't play nice with SSR. But I don't see why that couldn't be changed. Idk if handling SSR would make it different enough from Imperative Actions to warrant a different notation like `with:dropZone` as you mention.

There's a new RFC where your input would be greatly appreciated, you seem to have some nice thoughts on the matter. I honestly think it almost supersedes this one, but there's some discussion about it on that RFC.

The biggest drawback I see is:
You lose the ability to apply multiple actions (targeted slots in that RFC)  to the same element.

With declarative actions you can do:

```html
<div use:action1 use:action1 use:action2 use:action3>Hi!</div>
```

With targeted slots this doesn't really work. I do have some thoughts that could make this work with targeted slots but I'll brew them for some more time xD. I'll post them over there when I can formalize my thoughts.

https://github.com/sveltejs/rfcs/pull/68

EDIT: Ok... I completely overlooked that you (@mimbrown) had already read @lukaszpolowczyk's RFC. I won't change the comment though, just keep in mind I hadn't noticed xD

lukaszpolowczyk:
> With targeted slots this doesn't really work. I do have some thoughts that could make this work with targeted slots but I'll brew them for some more time xD. I'll post them over there when I can formalize my thoughts.

@Zizico2 I'm waiting for your solution to this problem, as I myself don't have any good idea about it yet.

mimbrown:
> It's funny you mention the material ripple. That was the first time I felt the need for something like Declarative Actions adn then I generalized from there. And I don't think you would need to add an element to implement the material ripple. I think adding elements would make this RFC more convoluted than it needs to be and we would lose the differentiation between vertical and horizontal that @Prinzhorn conjectured in an earlier comment.

It is interesting we had the same use case in mind! I think there are inevitably going to be use cases where adding child elements through a declarative action will be helpful/necessary, but I agree that it complicates the proposal. I would be in favor of a V1 where children are disallowed and continuing this discussion for future versions.

> You're right that using the current action infrastructure maybe wouldn't play nice with SSR. But I don't see why that couldn't be changed. Idk if handling SSR would make it different enough from Imperative Actions to warrant a different notation like with:dropZone as you mention.

About SSR, I don't think there's a scenario where the current `use:action` syntax will support SSR in the future. You can't really tell users, "An action is a function that takes a DOM element and performs custom logic on it, except you have to support the scenario where we are in an SSR context and your function is actually getting passed a pseudo non-DOM element." That was the main gist I got from #4375, and I agree. Since actions already exist as a non-compiled feature, I think something different might be required. That's my 2 cents.

> EDIT: Ok... I completely overlooked that you (@mimbrown) had already read @lukaszpolowczyk's RFC. I won't change the comment though, just keep in mind I hadn't noticed xD

Haha np. It goes without saying that I agree with my comments over there :)

PR #44, opened 2020-12-13
Local Components

pushkine:
[Click here to read the RFC](https://github.com/pushkine/rfcs/blob/patch-1/text/0000-local-components.md) 
_please do not comment hot takes unless you've read the RFC 😄, below is just a TL;DR_

Components often require their own private building blocks, 
unfortunately in Svelte you must create a new file to declare a component.

This RFC proposes for `.svelte` files to define named components after the default export 
defining a local component is done at the top level through a non-indentating syntax such as `<!-- LocalComponentName -->`

Eq0gHiPW8AEom8T

### DIFF
--- a/text/0000-local-components.md
+++ b/text/0000-local-components.md
@@ -16,6 +16,7 @@ Each Component requires creating a brand new `.svelte` file, but sometimes a Com
 3. a node uses completely different attributes/directives depending on a condition
 4. a variable from a `let:` slot prop, an `#await` resolved promise or a `#each` value is transformed then used more than once
 5. an `#each` block needs its own script for `$`subscriptions or lifecycle
+6. a Component is too large and requires to be broken down for readability sake
 
 Having to spread tiny parts of a Component across several files is unfortunate as it leads to duplicate component names, cluttered folders and quite the refactoring hazard
 
--- a/text/0000-local-components.md
+++ b/text/0000-local-components.md
@@ -1,4 +1,4 @@
-- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- Start Date: 2020-12-13
 - RFC PR: ([#44](https://github.com/sveltejs/rfcs/pull/44))
 - Svelte Issue: (leave this empty)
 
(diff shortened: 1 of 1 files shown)


Monkatraz:
A bit of syntax that Svelte sorta has already (and that isn't easily found in the documentation) is this:
```svelte
<!-- @component JSDoc-esque component can go here. -->
<script/>
<style/>
<content/>
```
IMO, I think using a comment is a bit... vague? I feel like it could cause edge cases. Instead, I would prefer something like this:
```svelte
{@component component}
<script/>
<style/>
<content/>
```
This keeps the intuition of the `@component` token for documentation comments.

Alternatively, you could use the proposed `<svelte:template>` syntax and instead nest the local component inside of a tag rather than using a comment or `{...}` mechanism.

stshepana-eng (approved):
Scho

Crisfole:
I was coming to find out if someone had an idea like this. I agree with the reasoning, even if I'm not 100% convinced it's strictly needed. It could make demos more concise too. I dislike using comments semantically, though I think we already have a solution built in, just need to repurpose it:

```html
<script context="module">/* ... */</script>

<!-- This is a component accessible as <Foo> _inside_ this file only -->
<svelte:component name=Foo><!-- ... --></svelte:component>

<!-- This is a component accessible as <Bar> _inside_ this file only. It is exported by name as well. -->
<svelte:component export name=Bar><!-- ... --></svelte:component>

<!-- Any other content _outside_ the svelte:component remains the default export -->
<script>
  // Blah blah.
</script>

<Foo /> <!-- Didn't need to be imported since it's defined in-file -->
<Bar /> <!-- Didn't need to be imported since it's defined in-file -->
```

pepa-linha:
One file, one component. It is simple, clean and prepare to reusable.

samclaus:
Yeah I caught myself thinking about small one-off modal templates today and wondering if frameworks (not just Svelte) should support inline/local components. Then I stopped myself.

1. The "exportability" concept has a smell. If you're just gonna export them anyways, may as well give them their own file the old-fashioned way.
2. Small, one-off things have a tendency to grow into bigger, needs-to-be-reusable things (in my experience).
3. I write Angular on the job and the only reason I've ever preferred templates over full-fledged components was because Angular components ALWAYS wrap their markup in a host element at runtime, so you cannot have recursive components that produce a flat DOM list, etc. Svelte components do not have this problem to start with.

Honestly this one seems like an antifeature to me, unless people can provide several examples where it's really useful. Seems like it would take quite a bit of work to implement.

PR #47, opened 2021-03-29
Added definition for RFC inline with first mention

benaltair:
It was unclear to a novice what RFC meant.

Simply added `Request for Comments (RFC)` inline in `README.md` to expand the acronym at the first mention of an RFC in the paragraph text.


benaltair:
Alternatively something like `"RFC" (request for comments)` could be added like on:
https://github.com/reactjs/rfcs/blob/master/README.md

dummdidumm (approved):
Thanks!

PR #48, opened 2021-04-10
return:condition directive

MarkTanashchuk:
[rendered](https://github.com/sveltejs/rfcs/blob/5b1cd03c791304ae5832be4012aa8c047765e5e0/0000-return-directive.md)


MarkTanashchuk:
I apologize in advance for my poor English.
This is my first rfc, so some inaccuracies are also possible.

dkzlv:
This looks too specific to a very limited number of use cases. I've read your RFC carefully and still I hardly understand what you're proposing, but it seems most of this can be solved by spreading attributes into the tag:

```svelte
<script>
  $: stuff = {
    // Combine it from static stuff and props
  }
</script>

<input {...stuff} />
```

Though it's not an ideal solution (no support for `bind:` and `on:` directives), it totally your core issue of repeating the markup.

pngwn:
Thanks for taking the time to create an RFC!

I think the biggest issue with this is it is a conceptual break from how element attributes and component properties  work, which makes the semantics very confusing.

Semantics for component properties, component directives and element attributes are broadly the same from a semantic point of view so I'll just refer to "props" from now on but what I'm about to say holds true for all of these constructs. When we add a prop to a component or element we can guarantee that it will be applied regardless of the order in which it is applied. Broadly speaking it doesn't matter what that order is, they will still be applied (there is a caveat to this but it is an exception). Because of this, this proposal would be a breaking change of the most confusing kind: it would completely change the semantics of props without changing any syntax. The introduction of this single directive would completely rewrite the book on how props are applied.

> *I mentioned there is a caveat regarding order: `on:` and `use:` directives will always apply regardless of order, as mentioned above, but they are applied in the order that they are listed because they both interact directly with the node itself and this can have consequences. This is a feature that was added to solve a specific problem and is very useful if you need it. So there is one situation in which order is significant, although not as significant as in this proposal.*

To illustrate this point more clearly, it took me several readings of the RFC to even understand what what this new directive would do. While early return semantics are quite clear to me in JS/TS when I switch to a different context (XML) my brain flips a switch because the semantics are entirely different and I have become conditioned to make that 'switch' without even thinking about it. While we have sometimes added new semantics on top of languages or modified them slightly (`export let`, `$: x = y`, etc.), we try to avoid design that is completely at odds with existing semantics. I feel that this proposal falls into that category.

There is definitely some value in this proposal as illustrated by your examples but I think there is a very strong case that the `if/else` is clearer and more inline with the intention (it is also more easily optimised by the svelte compiler). We have always been strongly against control flow directives due to the fact that the visual landmarking is so poor. While this proposal sits somewhere in between, it also suffers from the same problem. As explained by @dkzlv this can also be partially solved with spread props (sans directives).

I'd propose that better component design could address most of these issues though. By creating more components and composing them aggressively, you could share a huge amount of the shared styling and logic making those parts that need to different as additional component to be composed. This could shrink bundles, simplify the code, reduce API surface area (making breaking changes less likely), and give the consumer more flexibility. Something like this:

```svelte
<Input.Wrap>
  <Input.Number />
</Input.Wrap>

<Input.Wrap>
  <Input.Text />
</Input.Wrap>
```

You could also re-export these composed components as separate named components. Due to the fact that they are both using the same component wrapper (which is a distinct module), bundlers will optimise for this case and make sure it is only bundle in your application once and shared with all consumers.

I feel like a good component design preferring composition will help to solve this problem in a clearer, more performant, and more easily maintainable way. Making use of slots, the context API, and stores will help with the implementation.

---

There are other solutions to this problem and the best place to solve it is in the design of the component rather than by adding new API. Due to the fact that it would be a considerable breaking change of a very confusing nature and only offers a small amount of additional functionality, I do not think this is something that we will want to implement. 

I will leave this RFC open for further discussion before the maintainers make a final decision.

MarkTanashchuk:
> While early return semantics are quite clear to me in JS/TS when I switch to a different context (XML) my brain flips a switch because the semantics are entirely different

If I understand correctly, `{#await}`, `{#each}` and actions also require a similar approach and it's not much of a problem.
Please, correct me if I misunderstood your point.

> Though it's not an ideal solution (no support for bind: and on: directives), it totally your core issue of repeating the markup.<br />
> As explained by @dkzlv this can also be partially solved with spread props (sans directives).

Mainly because of directives, I would like something like this to be added.

Common attributes can be combined without much difficulties, but things get complicated when you need to group actions, listeners, transitions (transition:scale for one group, transition:fly for another, etc.) in addition to other common attributes.

I [updated repl](https://svelte.dev/repl/b90be4c311f04cac9736d420743199ad?version=3.37.0) and added a fourth version of the original if/else shortening.
The result (in my opinion) did not go far from the 2nd and 3rd options.

MarkTanashchuk:
> There is definitely some value in this proposal as illustrated by your examples but I think there is a very strong case that the if/else is clearer and more inline with the intention

After reading your message I took a fresh look at the Rfc and understood that this design can lead to confusion and that it does make sense to look for a design that is easier to adapt to more situations

Just duplicating markup in if/else and having though large but readable components is, actually, not a big problem.

But I think that development would be easier if there was some api that could help get rid of this kind of duplication(as described in the rfc example with if/else), and I would be happy if such a thing was added, in whatever form it might be.

> I'd propose that better component design could address most of these issues though. ... . Something like this:

```svelte
<Input.Wrap>
  <Input.Number />
</Input.Wrap>

<Input.Wrap>
  <Input.Text />
</Input.Wrap>
```

While writing this rfc I also had the idea of making certain <svelte:* /> elements that could be used in the component that could act as attribute grouping, but I didn't know if this kind of component change as you showed is appropriate.

If that's appropriate, what do you think about adding this kind of syntax?

> The only names I can think of which seem normal to me are `trait` and `struct`, even though in other languages they mean other things, like objects or sets of methods.

```svelte
  <svelte:trait
    this="text input"
    
    type="text"
    placeholder={"Text input"}
  />

  <svelte:trait
    this="number input"
    
    type="number"
    use:action
    bind:value={$someStore[property]}
  />
  
  <input
    trait:Text="text input"
    trait:Number="number input"
   />
```
And in a parent:
```svelte
<Input.Text />
<Input.Number />
```

(as I see it)
This is basically just a preprocessor that inserts attributes and directives on the tags before compilation.
This approach doesn't have problems with clarity like `return:` directive does, but it is supposed to be done only once at compile time, not as (could be) done with `return:` directive - when condition changes

pngwn:
> If I understand correctly, {#await}, {#each} and actions also require a similar approach and it's not much of a problem.
Please, correct me if I misunderstood your point.

They are blocks and while they are custom to Svelte, they have very clear semantics. What are are suggesting is breaking with how _attributes_ work. The semantics of attributes (which props and directives mimic) is such that they will _always_ be added regardless of order. That is what this proposal would change, and this is a fundamental concept to XML which I do not think we should be breaking.

pngwn:
Why not just create those `traits` as actual components? It is basically the same amount of code, you can share the share attributes via spread props.

MarkTanashchuk:
> Why not just create those `traits` as actual components?

Because in the case described in the repl, all `traits` have shared styles (e.g. input) and the only way to combine styles in multiple components (in this case) is to write global styles in one of them, which breaks isolation.

MarkTanashchuk:
Otherwise it's needed to duplicate both styles and `export let`'s with all js code that may belong to several components at the same time.

pngwn:
There are a number of ways to share design and behaviour across multiple components. I'd encourage you to come and discuss this in the Svelte Discord channel and our community will be happy to discuss some of the possibilities. 

I think this is the core of my issue, I have written many components that share both styles and behaviour whilst exposing different variants by leveraging component composition. Svelte has a number of features to support this use-case. A new API is not required.

Conduitry:
`return:` as something that attempts to conjure up `if(condition) return;` only really works when a component is first being constructed, and not when it needs to be updated later. We're not just running from the top, constructing a VDOM, bailing when we get to a `return`, and then diffing that with the DOM.

This seems to be a strangely syntax solving an arbitrary problem in a way that attempts to create a leaky abstraction, except the underlying thing that's leaking isn't really there.

MarkTanashchuk:
> This seems to be a strangely specific syntax solving an arbitrary problem in a way that attempts to create a leaky abstraction, except the underlying thing that's leaking isn't really there.

Actually, the only thing I wanted was to be able to group attributes and, especially, svelte's directives.

Whenever I created a component responsible for styling one tag (and all its sub-tags) and its directives, I had to write a long `{#if}{:else}{/if}`, in which many attributes and directives are often repeated in whole groups.

The larger such components, the more difficult they are to maintain, and combining their attributes and directives so that the component fits on one or two screens in height is not a particularly interesting task, but necessary if the markup is not divided into separate components every time component is extended.

Because if you move each implementation into a separate component, then you may end up with a folder with a dozen components, each of which deals only with managing directives/attributes and styling one tag (input and drop-down list are the simplest and most common examples).

> I'd encourage you to come and discuss this in the Svelte Discord channel and our community will be happy to discuss some of the possibilities.

I will look through my previous projects to find examples where disadvantages of `{#if}{:else}{/if}` and splitting it into separate components will be more noticeable and clear, and I will definitely start a discussion about them in discord.

MarkTanashchuk:
I think the topic of `return:` directives can be closed - a more flexible implementation is needed

PR #52, opened 2021-05-29
Abstract prerendering

johnnysprinkles:
[Rendered](https://github.com/johnnysprinkles/rfcs/blob/abstract_prerender/text/0000-abstract-prerender.md)

### DIFF
--- a/text/0000-abstract-prerender.md
+++ b/text/0000-abstract-prerender.md
@@ -1,0 +1,84 @@
+- Start Date: 2021-05-29
+- RFC PR: 
+- Svelte Issue: 
+
+# Abstract pre-render
+
+## Summary
+
+Adding support for prerendered HTML files which look for their page store data at client-side runtime (what I'd call "abstract"), instead of using the snapshot of such data as it was at build time (what I'd call "concrete").
+
+## Motivation
+
+This will allow us to pre-render every page, even the ones that do something with the request-specific page store data in `onMount`. It is possible to do this today with no changes to SvelteKit itself, if you have a web server that can read and transform (interpolating values) into the html artifacts for each request. But doing that is messy.
+
+There's a lot more detail about motivation in the initial comment of https://github.com/sveltejs/kit/issues/1533.
+
+## Detailed design
+
+### Technical Background
+
+Right now, a pre-rendered page has code like this to hydrate (this is for a parameterized route):
+
+    page: {
+      host: location.host, // TODO this is redundant
+      path: "/my/page/123",
+      query: new URLSearchParams("foo=bar"),
+      params: {"id": "123"}
+    }
+
+Aside from the "host", those are all the concrete, specific values as they were at build time, baked into the page as literals. But one might want to use the prerendered HTML artifact in an abstract way, for variable and wildcard request parameters, in which case you'd want something more like this:
+
+    page: {
+      host: location.host,
+      path: location.pathname,
+      query: new URLSearchParams(location.search),
+      params: parsePathParams(location.pathname, /^\/my\/page\/(\w)+$/, ["id"]),
+    }
+
+Where `parsePathParams` is a simple function that gives you an object of path params.
+
+### Implementation
+
+The current state of the code is here:
+
+https://github.com/sveltejs/kit/blob/a66affec4168fae053fb6349805b60028306abcf/packages/kit/src/runtime/server/page/render.js#L136
+
+The changes would be:
+
+* For host, remove the TODO comment and the ternary, and just always use `location.host`
+* For path, use `location.pathname`
+* For query, use `location.search`
+* Params is the hardest one, we'd need to insert the route regex and the positional list of parameter names into a simple `parsePathParams` function, possibly just inlined as an arrow function
+* Update the docs where appropriate
+
+Note that there is an in progress PR for the query part in https://github.com/sveltejs/kit/pull/1511 but in my view it would be better to address these all together in one PR.
+
+If we implemented this RFC, it fixes https://github.com/sveltejs/kit/issues/669 and https://github.com/sveltejs/kit/issues/1397, and partially fixes https://github.com/sveltejs/kit/issues/1561 and https://github.com/sveltejs/kit/issues/1533. This RFC is only about the page store hydration aspect but "Abstract Pre-rendering" has some other facets to it, discussed in https://github.com/sveltejs/kit/issues/1533, including the question of what values you provide for parameterized route params when adding paths to `kit.prerender.pages`.
+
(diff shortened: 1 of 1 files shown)


text/0000-abstract-prerender.md
```diff
@@ -0,0 +1,84 @@
...
+
+The changes would be:
+
+* For host, remove the TODO comment and the ternary, and just always use `location.host`
+* For path, use `location.pathname`
+* For query, use `location.search`
+* Params is the hardest one, we'd need to insert the route regex and the positional list of parameter names into a simple `parsePathParams` function, possibly just inlined as an arrow function
+* Update the docs where appropriate
+
+Note that there is an in progress PR for the query part in https://github.com/sveltejs/kit/pull/1511 but in my view it would be better to address these all together in one PR.
```
benmccann, line 55: This line is a bit outdated - the PR was closed

johnnysprinkles:
Here's how this would look, not submitting for PR or anything but just wanted to demonstrate the changes we'd need to make. Not very deep changes:

https://github.com/johnnysprinkles/kit/commit/963f0471184c245799744de7befa4ac85e82c56d

I think something like this would ultimately be a prerequisite for doing i18n as well. We'd ultimately need a prerenderer that can render each route multiple times based on an arbitrary number of dimensions, and presumably we'd pass information about which value for each dimension via query string. That communication is going to be different at runtime and at build time, for example the current location at runtime might just be geolocated and not use the query string at all. I'll post more thoughts in the i18n mega thread.

johnnysprinkles:
I really think we should reconsider the discussion in https://github.com/sveltejs/kit/pull/1511 that got shut down abruptly by a @Rich-Harris comment (https://github.com/sveltejs/kit/pull/1511#issuecomment-850907659). cc @pz-mxu

It's quite useful to be able to pre-render pages for which params aren't known at build time, and I certainly plan on doing just that.

johnnysprinkles:
Another possible approach would be do not do any of this, and let client-only code that needs to know query or route params simply not use the page store and get it directly from `location`. The only issue there is the client doesn't know the regex pattern or positional list of param names. We could just put that somewhere where the client code could get it.

johnnysprinkles:
Oh what am I thinking? runtime/client/rendered already knows how to parse the URL according to the current route. All I'd need to do is make it re-run its page store generation logic at the appropriate time, to overwrite the wildcard one that was captured in the page store from build time. Let me investigate this path.

johnnysprinkles:
So just to summarize, I have two approaches to demonstrate and a couple others to consider:

1. Redefine what the page store means, when pre-rendering is involved. Currently it means "the values as they were at pre-render time." This proposes that they become "the values as they currently are for the current request at runtime." Might be rather invasive to redefine things like this though. https://github.com/johnnysprinkles/kit/commit/963f0471184c245799744de7befa4ac85e82c56d
2. Make the primitives of the SSRRoute data available in the client somehow, maybe via singletons and $app/navigation, so they can parse the window.location themselves to get path params. Fewer changes in server/page/render.js https://github.com/johnnysprinkles/kit/commit/e83bd2151d55a0561c260b073358d4490067bf09
3. Maybe just add a simple hack that navigates to the current page, to get the page store to update? The downside of this is now we depend on the client router, which I wasn't planning to use. This seems like a non-starter.
4. Add two more fields into the page store, "routePattern" and "routeParamNames". For people who use client-side routing, this should update at you navigate around, even though I can only see it being useful on the first page. Now you can look at the route param names and if any seem dynamic (have colons, square brackets, or whatever convention you want to use), you can parse the path params from the URL yourself. This seems like the most correct way to go.

text/0000-abstract-prerender.md
```diff
@@ -0,0 +1,84 @@
...
+The current state of the code is here:
+
+https://github.com/sveltejs/kit/blob/a66affec4168fae053fb6349805b60028306abcf/packages/kit/src/runtime/server/page/render.js#L136
+
+The changes would be:
+
+* For host, remove the TODO comment and the ternary, and just always use `location.host`
+* For path, use `location.pathname`
+* For query, use `location.search`
+* Params is the hardest one, we'd need to insert the route regex and the positional list of parameter names into a simple `parsePathParams` function, possibly just inlined as an arrow function
```
benmccann, line 52: I'd proposed moving the parsing from built time to the runtime in https://github.com/sveltejs/rfcs/pull/36

text/0000-abstract-prerender.md
```diff
@@ -0,0 +1,84 @@
...
+    page: {
+      host: location.host, // TODO this is redundant
+      path: "/my/page/123",
+      query: new URLSearchParams("foo=bar"),
+      params: {"id": "123"}
+    }
+
+Aside from the "host", those are all the concrete, specific values as they were at build time, baked into the page as literals. But one might want to use the prerendered HTML artifact in an abstract way, for variable and wildcard request parameters, in which case you'd want something more like this:
+
+    page: {
```
benmccann, line 32: You don't have to use the page store. You can just use `location` on the client-side. I don't think we'd want the page store on the client to be different from the page store on the server

johnnysprinkles: Sure, for query strings you could circumvent the page store and look at location.search, but if you're using ssr and client routing, doesn't that make a weird situation where, on initial mount you have to use location but on subsequent pages you can use the page store? Seems more consistent if they can just always use the page store.

Also, for path params, just using location wouldn't be an option because the client doesn't have the pattern or param names.

text/0000-abstract-prerender.md
johnnysprinkles: Thanks, hadn't seen that before, let me look.

text/0000-abstract-prerender.md
benmccann: The query string and host aren't available at prerender time. You have to get them from the `location`. It's better to make this explicit

johnnysprinkles: That's fine, it's just that pesky path params then. I wonder if we could monkey patch a "params" property onto the `window.location` object. I mean of course we could, I wonder if you think we should? That would allow "location" to give a complete, dynamic view of everything the page store would normally contain. It's kind of abusing the Location builtin, and people who use Typescript might have trouble accessing a property that's not supposed to be on Location, but aside from that seems pretty elegant to me.

johnnysprinkles: Though, I guess this wouldn't work for people using client routing and static ssr. They'd have to look at `location.params` on the first page, and `$page.store.params` on subsequent pages.

johnnysprinkles: Regarding https://github.com/sveltejs/rfcs/pull/36, maybe that's the answer. If you need to know the path params from `onMount` in a static pre-rendered page (at client runtime), instead of page store or location, maybe it's `routerInstance.current().params` or similar. Just want to make sure this all works even for people not using client-side transitions, e.g. `router: false` in the Svelte config.

PR #54, opened 2021-07-05
Syntax sugar for reactive declarations

Mlocik97:
[Rendered](https://github.com/Mlocik97/rfcs/blob/master/text/0007-reactive-sugar.md)

### DIFF
--- a/text/0007-reactive-sugar.md
+++ b/text/0007-reactive-sugar.md
@@ -56,7 +56,7 @@ One of the drawback is that IDEs and text editors like VS Code `mark` variables 
 
 ## Alternatives
 
-I also considered to design with writing array of variables, that would be dependencies for reactive statement. The problem is, it's more bloated, and require writing variable name two times. I think `_` (prefix) idea is most ideal.
+I also considered to design with writing array of variables, that would be dependencies for reactive statement. The problem is, it's more bloated, and require writing variable name two times. I think `_` (prefix) idea is most ideal. Alternative for prefix can be `#`.
 
 ## Unresolved questions
 
--- a/text/0007-reactive-sugar.md
+++ b/text/0007-reactive-sugar.md
@@ -1,0 +1,63 @@
+- Start Date: 2021-07-04
+- RFC PR: 
+- Svelte Issue: https://github.com/sveltejs/svelte/issues/6488
+
+# Syntax sugar for reactive declarations
+
+## Summary
+
+This RFC propose breaking change, that would simplify usage of reactive assignments, when we use both variables that should be dependency for reactivity and those, that shouldn't be.
+
+## Motivation
+
+Today, when we need to create reactive declaration with function, that contain two parameters, and we want to `react` only, when one of them change, we need to encapsulate it to another function (`callmyfunction` in our examples), and write variable name two times. 
+
+```js
+/* thanks to Idle for example code */
+function callmyfunction() {
+  myfunction(callfuncwhenthisvarchange, dontcallwhenchange)
+}
+$: callfuncwhenthisvarchange, callmyfunction()
+
+```
+
+or
+
+```js
+/* thanks to Rainlife for example code */
+const callmyfunction = () => myfunction(callfuncwhenthisvarchange, dontcallwhenchange);
+$: callfuncwhenthisvarchange, callmyfunction();
+```
+
+that is a lot of unnecessary code...
+
+I suggest to add new feature / syntax sugar, that will not use variable for reactive assignments, if it will be prefixed with `_`, so previous code would be possible to rewrite to something like this:
+
+```js
+$: myfunction(callfuncwhenthisvarchange, _dontcallwhenchange);
+```
+
+Same can be aplied on expressions like:
+
+```js
+$: z = _x + y
+```
+
+and in other situations as well.
+
+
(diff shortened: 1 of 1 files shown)


Mlocik97:
I was thinking about this again (and create note for my self, You can ignore these texts for now), and it would require:


1. checking if at least one variable in reactive statement is "used", that's not problem.
2. prefixing stores will be `_$myvar`, that's not problem.
3. how to solve this problem:

```js
let x = 5;
let _x = 7;
let y = 4;
$: z = _x + y;
```

(where `_x` is referencing `x`, or should it reference `_x`? not to mentoy we can have defined both `x` and `_x` in one scope)

We probably can:

1. if `_x` is defined, we "unuse" it, if not, we "unuse" `x`. // I think this is bad, as it would lead to "unexpected" and ugly behavior.
2. always make `_x` reference to `x`, and `_x` variable be unable to be used in reactive statement. // I like this solution
3. throw error // little agressive
4. making `_variable` names "invalid" for Svelte, as it has special meaning for it. // I don't want this, it would go much against JS.

I think solution 2 is really good.

So:

`$: z = _x + y` -> `_x` here will always reference `x` variable, that means, in result code in example will result to that, z would be equal to 9, not 11, when y = 4.

text/0007-reactive-sugar.md
```diff
@@ -0,0 +1,63 @@
...
+const callmyfunction = () => myfunction(callfuncwhenthisvarchange, dontcallwhenchange);
+$: callfuncwhenthisvarchange, callmyfunction();
+```
+
+that is a lot of unnecessary code...
+
+I suggest to add new feature / syntax sugar, that will not use variable for reactive assignments, if it will be prefixed with `_`, so previous code would be possible to rewrite to something like this:
+
+```js
+$: myfunction(callfuncwhenthisvarchange, _dontcallwhenchange);
```
coyotte508, line 37: I think you should show a whole code sample including the declaration of `_dontcallwhenchange`, here we don't know if the `_` is on the variable name or just in the reactive statement

coyotte508: you should also specify the syntax when interacting with store content, is it `_$dontcallwhenchange` ?

Mlocik97: yes, as I wrote in comment... about declaration, it still something, we need to think about, but probably it would be good to have it only as prefix in reactive statement, means after compiled, prefix would be removed from statement, to match variable name without prefix. The only issue is, if You write callback in reactive statement, where `_` prefix is used naturally. Hmm... will think about it.

coyotte508: Then maybe update the RFC with that? :)

7nik:
what about using _control comments_?
something like
```js
let a = 1, b = 2, c = 3, d;

// do not trigger for a and b
$: { // svelte:skip a, b
  d = a + b + c; 
}

// trigger only for  b
$: { // svelte:only b
  d = a + b + c;
}

// and shortening
$: /* svelte:only c */ z = a*a + b*b + c*c;
```

Mlocik97:
using comments to affects what code is produced by transpiler seems like big "MEH"... not to mentoy it's too noisy, too much characters to write. Not even better than React's useEffects...

text/0007-reactive-sugar.md
Mlocik97: I was thinking more about, and what about `#` prefix?

7nik:
I agree, I haven't seen the usage of comments to control compilers, only for warnings. But this way has a good readability and is quite clear.

Another thing I was thinking is whitelists and blacklists:
```js
let a = 1, b = 2, c = 3, d;
// whitelist
$: (a,b), d = a + b + c;
// blacklist
$: !(c), d = a + b +c;
```
But here labels work only with the comma operator so this approach doesn't fit the reactive block of code.

But what if we use a lambda function to define reactive dependencies:
```js
let a = 1, b = 2, c = 3, d;
// re-calculated d only when a or b changes
$: (a,b) => {
  d = a + b + c;
}
// do not re-calculated d when c changes
$: !(c) => {
  d = a + b + c;
}
```

gtm-nayan:
Just gonna leave this here https://github.com/sveltejs/svelte/issues/6730#issuecomment-923714936 👀

lukaszpolowczyk:
Such a random idea:
```svelte
<script>
let y=1;
let z=2;
let a=3;
$(y,z): x = y + z + a;
</script>
```
...where overwriting the `a` variable does nothing. It only does an overwrite of the `y` or `z` variable.

I know that the `$:` syntax refers to `label:`, but honestly it wouldn't make any difference to me, because the reference is tongue-in-cheek anyway.

---

### IMPORTANCE:

@gtm-nayan This looks cool, but it would break the code because someone might have used it in a different way before.
Alternatively, something like this, with, for example, a double dollar `$$`:
```javascript
$$: {y,z} {x = y + z + a}
```
To distinguish one from the other.

lukaszpolowczyk:
Now even a workaround using functions won't help. xD :P :D
Greetings #svelte-5-runes

dummdidumm:
Svelte 5 solves this with the runes API. Any dependencies are tracked at runtime, and you can untrack them using the `untrack` function. Therefore closing this - thank you for the proposal.

PR #55, opened 2021-07-06
Add iso language proposal

acewf:


0000-iso-language-code.md
```diff
@@ -0,0 +1,67 @@
...
+So something within the lines of
+
+```
+// app.html
+<html lang="%svelte.lang%">
+```
+
+```
+/// page-to-render.svelte
+<svelte:lang data-lang="{$isoLang}"/>
```
Mlocik97, line 52: it looks like an unnecessarily long entry,... maybe `<svelte:html lang={isoLang} />` without `data-` that is totally useless.

PR #57, opened 2021-07-30
Implement spread binding

sxxov:
[Rendered](https://github.com/Sxxov/rfcs/blob/master/text/0000-spread-binding.md).

### DIFF
--- a/text/0000-spread-binding.md
+++ b/text/0000-spread-binding.md
@@ -1,0 +1,404 @@
+- Start Date: 2021-7-30
+- RFC PR: (leave this empty)
+- Svelte Issue: (leave this empty)
+
+# Spread Binding
+
+## Summary
+
+This RFC proposes a way to bind all properties inside an object, similar to the existing [spread syntax](https://svelte.dev/tutorial/spread-props).
+
+```sv
+<Component bind:...={object} />
+```
+
+## Motivation
+
+The below motivators inherit any ones that spread props had, as it's similar in function.
+
+### Bind Forwarding
+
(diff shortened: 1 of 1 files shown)


Florian-Schoenherr:
really extensive 😄 
I think the `...` solution is pretty neat.

RiseNet-Web:
> really extensive smile
> I think the `...` solution is pretty neat.

Same

manvankaliber:
I would love to see this implemented.

ghost:
I like the idea, but I think a `bind={object}` syntax would be more terse.

0x6563:
Because the spread operator already exists for passing attributes, it seems tempting to use similar syntax for binding and even event handling. But I feel that semantically it doesn't really make sense.

To give you an example where I am coming from, a common practice when you use destructuring assignment in JS you do so to create a new object with a new reference, to prevent mutation on the source object. 

ie: `{ url: 'foo.bar', headers: {...headers, authorization: 'API TOKEN' }  }`.

As much I like your suggestion of `bind:{...object}` which is easy to interpret at first glance. I think a good alternative maybe to use `*` as a wildcard. 
ie:
`bind:*={object}`
`on:*={object}`
`*={object}`

Some Pros:
- `*` is common enough as a wildcard 
- Wont require introducing any new syntax or logic at the parser level See: [Ref](https://github.com/sveltejs/svelte/blob/c4261abfde96f10576b860ccab82951bf2cc35ef/src/compiler/parse/state/tag.ts#L299)
- Can be handled in the [Binding](https://github.com/sveltejs/svelte/blob/c4261abfde96f10576b860ccab82951bf2cc35ef/src/compiler/compile/nodes/Binding.ts#L32) and [EventHandler](https://github.com/sveltejs/svelte/blob/c4261abfde96f10576b860ccab82951bf2cc35ef/src/compiler/compile/nodes/EventHandler.ts#L11) classes similar to how [Attributes](https://github.com/sveltejs/svelte/blob/c4261abfde96f10576b860ccab82951bf2cc35ef/src/compiler/compile/nodes/Attribute.ts#L32) works.

simonhvc:
This would be a nice feature to have. It helps us create more dynamic components and wrappers. Any updates on this?

st3v3y:
Any updates on this? It would be definitely a nice feature to have.

mark-russ:
I hit a snag where I could really make use of this feature.

PR #59, opened 2021-09-29
feat: svelte config util

dominikg:
Please share additional ideas.

### DIFF
--- a/text/0000-svelte-config-util.md
+++ b/text/0000-svelte-config-util.md
@@ -9,7 +9,6 @@
 Introduce a new utility package `@sveltejs/config` to offer tools in the svelte ecosystem a common way
 to read and validate svelte config files and provide config features to users in a consistent manner.
 
-
 ## Motivation
 
 Currently, tools implement config reading separately (or not at all),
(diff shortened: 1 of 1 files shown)


dummdidumm:
I like this very much! About drawbacks: None of the listed are of concern to me, I think they force us to have a good talk about the config which is a good thing. I also don't think that the overhead is that big (bundlesize-wise).

I have one question about namespaces: I remember you mentioning that you could repeat top level entries in the namespace to overwrite the top level entries. Is that part of the solution? If yes, could you detail that part a little?

dominikg:
I had that added already but changed my mind. I think the inlineOptions in loadConfig give tools enough freedom

bluwy:
Regarding validation, I think it's of Rich's interest to re-use the code from Kit (also mentioned in RFC), and avoid a third-party library for it. I've made [a pull request](https://github.com/sveltejs/kit/pull/2366) in the past to convert to [superstruct](https://github.com/ianstormtaylor/superstruct), but was closed because the package size is quite hefty, and the error messages aren't really friendly.

In my experience refactoring that, making Kit's validation generic won't be easy unless we draw the line on the kind of validations we would only support. Otherwise, if someone can suggest an alternative mini validation library, would be nice too.

Or another route we could go is to dismiss validation all together and rely on types only.

benmccann:
Linking to past discussion about this for reference: https://github.com/sveltejs/vite-plugin-svelte/issues/146. There was a fair amount of discussion around what should be a top-level shared key vs living under a namespace for each bundler

dummdidumm:
One more question about transpiling TS->JS: would that affect the config file only, or would all transitive imports be transpiled as well? People will expect the latter I guess, but I'm not sure how this could be implemented right now.

bluwy:
> would that affect the config file only, or would all transitive imports be transpiled as well?

For `esbuild`, since it's a bundler we can bundle the config file and run it, similar to how [Vite works](https://github.com/vitejs/vite/blob/b17444fd97b02bc54410c8575e7d3cb25e4058c2/packages/vite/src/node/config.ts#L891). I'm not sure about `typescript` though, looks like Jest [uses ts-node](https://jestjs.io/docs/configuration) to load it. If we go down that path, we could as well rely on https://github.com/lukeed/tsm instead.

bluwy:
Maybe @DeMoorJasper from parcel-plugin-svelte could chime in too. IIRC parcel has its own config reading logic, plus Parcel had some big news coming soon (might affect this). So if it's not possible for them to use `@sveltejs/config`, perhaps it can be based on as a spec too.

dominikg:
leaving this here as a reminder: https://github.com/sveltejs/kit/pull/3219

add error handling for missing config file including links to documentation how to add/use svelte config

fzn0x:
> Had a chat with @bluwy about this, takeaways:
> 
> * types for common namespaces like preprocess / compilerOptions etc can be provided by `@sveltejs/config` but in general tools are responsible for their own types
> * loading .ts files can be done similar to vite by lazy requiring esbuild, bundle .ts to .js on the fly and then using common .js loading
>   
>   * async config feature allows users to plug in any other loading mechanism they like, so if esbuild doesn't tick their box, that would be an escape hatch
> * validation rules are complex to define a common ground for, json schema is ugly, superstruct is a bigger dependency and was discarded in kit before ( [[chore] use superstruct for config validation kit#2366](https://github.com/sveltejs/kit/pull/2366) ) so to keep the scope of this RFC and the initial release of `@sveltejs/config` more manageable, it should be dropped and tools keep their own validations until a later feature release adds validation support

Hi @dominikg is this the final takeaways?

This RFC provides the guide for @sveltejs/config package which helps svelte kit to bundle/transpile `svelte.config.{js,ts,cjs,mjs}` on the fly. I believe the challenging part is to widely support "allow svelte.config.ts". Looks like lilconfig does not support TypeScript config as well.

For `* defined namespacing (top-level keys owned by a tool)` do we need to separate PRs for that? 

So the idea is to make @sveltejs/config support different bundlers with namespace?

text/0000-svelte-config-util.md
```diff
@@ -0,0 +1,165 @@
...
+//...
+const config = await load_config();
+```
+
+loadConfig accepts an optional `options` argument with the following options:
+```js
+options = {
+    fromDir: '/some/path', // process.cwd by default, tools can pass what they need, eg vite root
+    inlineConfig: {/*...*/}, // svelte config object built from inline options eg cli args passed to the tool. merged in
+    namespace: 'kit', // namespace for the tool loading the config, useful for advanced features
```
fzn0x, line 69: Could you elaborate this one?

text/0000-svelte-config-util.md
```diff
@@ -0,0 +1,165 @@
...
+
+## How we teach this
+
+### To tool authors
+
+README.md and/or docs/ in the new GitHub repo, example usage in kit and vite-plugin-svelte
+
+### To users
+
+Currently, svelte.config.js isn't really documented on its own. The tools show how they use it.
```
fzn0x, line 135: maybe just add README.md in `@sveltejs/config` packages folder

text/0000-svelte-config-util.md
fzn0x: I mean this one

```diff
+      namespace: 'kit', // namespace for the tool loading the config, useful for advanced features
```

dominikg: if sveltekit calls `@sveltejs/config`  to load the svelte config, it can identify itself which enables sveltejs/config to implement some kit specific features or just have better logging.

text/0000-svelte-config-util.md
dominikg: That would be hard to keep in sync for the different namespaces where the actual type depends on the version used (eg sveltekit3 is going to have config changes that need documenting)

We do have some ideas for improving documentation discoverability though

PR #60, opened 2021-10-07
Implement forward directive

tropicaaal:
[Rendered](https://github.com/Tropix126/rfcs/blob/master/text/0000-forward-directive.md).

### DIFF
--- a/text/0000-forward-directive.md
+++ b/text/0000-forward-directive.md
@@ -1,0 +1,97 @@
+- Start Date: 2021-08-07
+- RFC PR: (leave this empty)
+- Svelte Issue: (leave this empty)
+
+# The `forward` directive.
+
+## Summary
+
+This RFC proposes a new `forward` directive, which lets you forward directives used on a component instance to a DOM element in said component. This allows you to call directives such as `use`, `transition`, `animation`, etc... on a component declaration. It also allows you to delegate all events to a certain element in a component.
+
+## Motivation
+
+At the moment, there are a long list of pitfalls with community-provided component libraries. Designing a "one-size-fits-all" component in svelte can be rather tedious in the current enviornment, notably when dealing with directives. For example, if you wish to forward commonly used events to a reusable <Button> component, you'll have to retype a long list of events like so:
+
+```html
+<button
+    on:click
+    on:blur
+    on:focus
+    on:dblclick
+    on:contextmenu
+    on:mousedown
+    on:mouseup
+    on:mouseover
+    on:mouseout
+    on:mouseleave
+    on:keypress
+    on:keydown
+    on:keyup
+    class="button {customClass || ''}"
+    {...$$restProps}
+>
+    <slot />
+</button>
+```
+
+A [solution](https://github.com/sveltejs/svelte/pull/4599) to this event issue using the `on:*` directive has been proposed. This syntax doesn't come without issues either, as `*` is nonstandard syntax in the scope of HTML events, and it doesn't reach the core limitations of component directives.
+
+Another common problem that comes up is when working with actions. Components are incapable of utilizing the `use` directive, meaning library authors often need to export a prop like so:
+```html
+<script>
+    export let use = () => {};
+</script>
+
+<div use:use></div>
+```
+
+While this also is effective in solving the issue, it adds further boilerplate and nonstandard syntax that other developers are forced to work with. Similar problems arise with the `transition` and `animation` directives.
+
+## Detailed design
+
+### Forwarding all events to an element
+
+This example uses the same <Button> component described earlier.
+  
+```html
+<button forward:on class="button {customClass || ''}" {...$$restProps}>
+  <slot />
+</button>
+```
+
+## Forwarding actions and transitions
+
+The `forward` directive can be similarly used to forward actions and transitions using the native syntax:
+
+```html
+<button forward:use forward:transition>
+  <slot />
+</button>
+```
+  
+Using the above component in parent context would look like this:
+
+```html
+<Button use:myAction>I'm a button with an action!</Button>
+```
+
+### Implementation
+
+The `forward` directive's syntax would take in the name of the directive to be forwarded. For example, to forward all events using the `on` directive, the corresponding directive would be `forward:on`.
+
+## How we teach this
+
+The directive should be taught as a method of using directives such as `use` or `transition` through the parent context of a component. It should be taught as a relatively advanced feature towwards the end of the tutorial, mainly intended for package authors. It wouldn't majorly affect other documentation topics, although it might be worth noting on the [Event Forwarding](https://svelte.dev/tutorial/event-forwarding) topic of the tutorial.
+
+## Drawbacks
+
+- There are some directives that could not be forwarded, most notably `bind` and `class`.
+
+## Alternatives
+
+- Keep things the way they are.
+- Use the `on:*` or the `bubble` directive proposed in [#4599](https://github.com/sveltejs/svelte/pull/4599) could be used as a solution for the event problem. Component authors could continue using nonstandard syntax for delegating other events such as `transition`, `use`, or `animation`.
+
+## Unresolved questions
+
+- How should the `bind` directive be implemented in this context?
--- a/text/0000-forward-directive.md
+++ b/text/0000-forward-directive.md
@@ -10,7 +10,7 @@ This RFC proposes a new `forward` directive, which lets you forward directives u
 
 ## Motivation
 
-At the moment, there are a long list of pitfalls with community-provided component libraries. Designing a "one-size-fits-all" component in svelte can be rather tedious in the current enviornment, notably when dealing with directives. For example, if you wish to forward commonly used events to a reusable <Button> component, you'll have to retype a long list of events like so:
+At the moment, there are a long list of pitfalls with community-provided component libraries. Designing a "one-size-fits-all" component in svelte can be rather tedious in the current enviornment, notably when dealing with directives. For example, if you wish to forward commonly used events to a reusable `<Button>` component, you'll have to retype a long list of events like so:
 
 ```html
 <button
@@ -51,7 +51,7 @@ While this also is effective in solving the issue, it adds further boilerplate a
 
 ### Forwarding all events to an element
 
-This example uses the same <Button> component described earlier.
+This example uses the same `<Button>` component described earlier.
   
 ```html
 <button forward:on class="button {customClass || ''}" {...$$restProps}>
--- a/text/0000-forward-directive.md
+++ b/text/0000-forward-directive.md
@@ -108,6 +108,27 @@ Similarly to the `on` forwarding, internal actions may be used along with the `f
 
 Note: A transition directive cannot be both forwarded and used internally, since [DOM elements can only have one transition](https://svelte.dev/repl/08236eb2bbc6474bad75d1aee43b2628).
 
+## Forwarding Style Directives
+
+The newly introduced [style](https://svelte.dev/docs#template-syntax-element-directives-style-property) directive can also be forwarded.
+
+```html
+<button forward:style>
+  <slot />
+</button>
+```
+
+Parent:
+```html
+<script>
+  import Button from "./Button.svelte";
+
+  let color = "red";
+</script>
+
+<Button style:color>I'm a red button!</Button>
+```
+
 ### Implementation
 
 The `forward` directive's syntax would take in the name of the directive to be forwarded. For example, to forward all events using the `on` directive, the corresponding directive would be `forward:on`.
--- a/text/0000-forward-directive.md
+++ b/text/0000-forward-directive.md
@@ -153,7 +153,7 @@ In this example, the `use` directive would be both forwarded to parent context a
 
 ## Potential Changes with Svelte 4
 
-It has been further proposed to completely replace the typical event forwarding syntax in the past for Svelte 4, this additional proposal would also cover that:
+It has been further proposed to completely replace the typical event forwarding syntax in the past for Svelte 4. This additional proposal could also cover that:
 
 ```html
 <button on:click></button>
--- a/text/0000-forward-directive.md
+++ b/text/0000-forward-directive.md
@@ -99,12 +99,14 @@ Using the above component in parent context would look like this:
 <Button use:myAction>I'm a button with an action!</Button>
 ```
 
-Similarly to the `on` forwarding, internal actions and transitions may be used along with the `forward` directive:
+Similarly to the `on` forwarding, internal actions may be used along with the `forward` directive:
 ```
-<button forward:use forward:transition transition:internalTransition use:internalAction>
+<button forward:use use:internalAction>
   <slot />
 </button>
 ```
+
+Note: A transition directive cannot be both forwarded and used internally, since [DOM elements can only have one transition](https://svelte.dev/repl/08236eb2bbc6474bad75d1aee43b2628).
 
 ### Implementation
 
@@ -116,7 +118,8 @@ The directive should be taught as a method of using directives such as `use` or 
 
 ## Drawbacks
 
-- There are some directives that could not be forwarded, most notably `bind` and `class`.
+- There are some directives that could not be forwarded, most notably `bind`.
+- You cannot combine `forward:transition` and `transition:fn` on the same element, since *[An element can only have one 'transition' directive](https://svelte.dev/repl/08236eb2bbc6474bad75d1aee43b2628)*.
 
 ## Alternatives
 
@@ -125,5 +128,5 @@ The directive should be taught as a method of using directives such as `use` or 
 
 ## Unresolved questions
 
-- How should the `bind` directive be implemented in this context?
-- How should svelte handle multiple transitions at once? Using an "internal transition" along with forwarding might result in some unexpected behavior.
+- Should the `bind` directive be handled here at all?
+- Should there be a syntax to exclude certain types of a directive? e.g. *forward all events except ones with the name "keydown"*.
--- a/text/0000-forward-directive.md
+++ b/text/0000-forward-directive.md
@@ -81,7 +81,7 @@ The `forward` directive's syntax would take in the name of the directive to be f
 
 ## How we teach this
 
-The directive should be taught as a method of using directives such as `use` or `transition` through the parent context of a component. It should be taught as a relatively advanced feature towwards the end of the tutorial, mainly intended for package authors. It wouldn't majorly affect other documentation topics, although it might be worth noting on the [Event Forwarding](https://svelte.dev/tutorial/event-forwarding) topic of the tutorial.
+The directive should be taught as a method of using directives such as `use` or `transition` through the parent context of a component. It should be taught as a relatively advanced feature towards the end of the tutorial, mainly intended for package authors. It wouldn't majorly affect other documentation topics, although it might be worth noting on the [Event Forwarding](https://svelte.dev/tutorial/event-forwarding) topic of the tutorial.
 
 ## Drawbacks
 
@@ -95,9 +95,11 @@ The directive should be taught as a method of using directives such as `use` or 
 ## Unresolved questions
 
 - How should the `bind` directive be implemented in this context?
-- How should `forward:on` or `forward:use` interact with existing `on:` or `use` directives?
+- How should `forward` interact with existing `on`/`use`/`transition`/etc directives?
 
 ## Additional Proposal - `foward` effects and singular event forwarding
+
+This section provides a possible answer to the second unresolved question, as well as a discussion of how `forward` can be integrated into Svelte 4 and onward.
 
 Another common problem that people run into is adding side-effects to forwarded events. Suppose a developer had the aforementioned `<Button>` component that forwarded the `click` event, but wanted to execute their own code along with the parent's click callback. The go-to solution would be event dispatchers:
 
--- a/text/0000-forward-directive.md
+++ b/text/0000-forward-directive.md
@@ -39,20 +39,30 @@ A [solution](https://github.com/sveltejs/svelte/pull/4599) to this event issue u
 Another common problem that comes up is when working with actions. Components are incapable of utilizing the `use` directive, meaning library authors often need to export a prop like so:
 ```html
 <script>
-    export let use = () => {};
+	export let use = [];
+	
+	const [action, options] = use;
 </script>
 
-<div use:use></div>
+<button use:action={options}>
+	<slot />
+</button>
 ```
 
-While this also is effective in solving the issue, it adds further boilerplate and nonstandard syntax that other developers are forced to work with. Similar problems arise with the `transition` and `animation` directives.
+While this also is effective in solving the issue, it adds further boilerplate and nonstandard syntax that other developers are forced to work with.
+
+```html
+<Button use={[myAction, {option: true}]}>Hi</Button>
+```
+
+Similar problems arise with the `transition` and `animation` directives.
 
 ## Detailed design
 
 ### Forwarding all events to an element
 
-This example uses the same `<Button>` component described earlier.
-  
+This example uses the same `<Button />` component described earlier.
+
 ```html
 <button forward:on class="button {customClass || ''}" {...$$restProps}>
   <slot />
@@ -117,19 +127,3 @@ The directive should be taught as a method of using directives such as `use` or 
 
 - How should the `bind` directive be implemented in this context?
 - How should svelte handle multiple transitions at once? Using an "internal transition" along with forwarding might result in some unexpected behavior.
-
-## Potential Changes with Svelte 4
-
-It has been further proposed to completely replace the typical event forwarding syntax in the past for Svelte 4. This additional proposal could also cover that:
-
-```html
-<button on:click></button>
-```
-
-would be changed to
-
-```html
-<button forward:on:click></button>
-```
-
-This is a breaking change, and likely out-of-scope.
--- a/text/0000-forward-directive.md
+++ b/text/0000-forward-directive.md
@@ -6,7 +6,7 @@
 
 ## Summary
 
-This RFC proposes a new `forward` directive, which lets you forward directives used on a component instance to a DOM element in said component. This allows you to call directives such as `use`, `transition`, `animation`, etc... on a component declaration. It also allows you to delegate all events to a certain element in a component.
+This RFC proposes a new `forward` directive, which lets you forward directives used on a component instance to a DOM element in said component. This allows you to call directives such as `use`, `transition`, `animation`, etc... on a component declaration. It also allows you to delegate all events to a certain element in a component's template.
 
 ## Motivation
 
--- a/text/0000-forward-directive.md
+++ b/text/0000-forward-directive.md
@@ -59,6 +59,20 @@ This example uses the same `<Button>` component described earlier.
 </button>
 ```
 
+This version of the component has an internal click handler that will run along with the forwarded `click` event.
+
+```html
+<script>
+    function internalFunction() {
+        console.log("Button Clicked!");
+    }
+</script>
+
+<button forward:on on:click={internalFunction}" {...$$restProps}>
+  <slot />
+</button>
+```
+
 ## Forwarding actions and transitions
 
 The `forward` directive can be similarly used to forward actions and transitions using the native syntax:
@@ -73,6 +87,13 @@ Using the above component in parent context would look like this:
 
 ```html
 <Button use:myAction>I'm a button with an action!</Button>
+```
+
+Similarly to the `on` forwarding, internal actions and transitions may be used along with the `forward` directive:
+```
+<button forward:use forward:transition transition:internalTransition use:internalAction>
+  <slot />
+</button>
 ```
 
 ### Implementation
@@ -95,63 +116,7 @@ The directive should be taught as a method of using directives such as `use` or 
 ## Unresolved questions
 
 - How should the `bind` directive be implemented in this context?
-- How should `forward` interact with existing `on`/`use`/`transition`/etc directives?
-
-## Additional Proposal - `foward` effects and singular event forwarding
-
-This section provides a possible answer to the second unresolved question, as well as a discussion of how `forward` can be integrated into Svelte 4 and onward.
-
-Another common problem that people run into is adding side-effects to forwarded events. Suppose a developer had the aforementioned `<Button>` component that forwarded the `click` event, but wanted to execute their own code along with the parent's click callback. The go-to solution would be event dispatchers:
-
-```html
-<script>
-    import { createEventDispatcher } from "svelte";
-    
-    const dispatch = createEventDispatcher();
-    
-    function handleClick(event) {
-        console.log(event.target);
-    
-        dispatch("click");
-    }
-</script>
-
-<button on:click={handleClick}>
-    <slot />
-</button>
-```
-This solution has underlying issues:
-- Event dispatchers lose the properties of `MouseEvent`.
-- This is a lot of boilerplate to perform a simple action.
-
-Instead of using dispatchers to add side-effects to events, we can extend the forward event to bubble singular event types as well. This could be done like so:
-
-```html
-<script>    
-    function handleClick(event) {
-        console.log(event.target);
-    }
-</script>
-
-<button forward:on:click={handleClick}>
-    <slot />
-</button>
-```
-
-This pattern could be similarly applied to actions:
-
-```html
-
-<script>
-    function internalAction(node) {
-        // do something with `node`
-    }
-</script>
-
-<button forward:use:internalAction></button>
-```
-
-In this example, the `use` directive would be both forwarded to parent context as well as `internalAction` being applied. A possible pitfall would be how the animation API would play into this, as it's unclear how the compiler would deal with multiple animations.
+- How should svelte handle multiple transitions at once? Using an "internal transition" along with forwarding might result in some unexpected behavior.
 
 ## Potential Changes with Svelte 4
 
--- a/text/0000-forward-directive.md
+++ b/text/0000-forward-directive.md
@@ -95,3 +95,74 @@ The directive should be taught as a method of using directives such as `use` or 
 ## Unresolved questions
 
 - How should the `bind` directive be implemented in this context?
+- How should `forward:on` or `forward:use` interact with existing `on:` or `use` directives?
+
+## Additional Proposal - `foward` effects and singular event forwarding
+
+Another common problem that people run into is adding side-effects to forwarded events. Suppose a developer had the aforementioned `<Button>` component that forwarded the `click` event, but wanted to execute their own code along with the parent's click callback. The go-to solution would be event dispatchers:
+
+```html
+<script>
+    import { createEventDispatcher } from "svelte";
+    
+    const dispatch = createEventDispatcher();
+    
+    function handleClick(event) {
+        console.log(event.target);
+    
+        dispatch("click");
+    }
+</script>
+
+<button on:click={handleClick}>
+    <slot />
+</button>
+```
+This solution has underlying issues:
+- Event dispatchers lose the properties of `MouseEvent`.
+- This is a lot of boilerplate to perform a simple action.
+
+Instead of using dispatchers to add side-effects to events, we can extend the forward event to bubble singular event types as well. This could be done like so:
+
+```html
+<script>    
+    function handleClick(event) {
+        console.log(event.target);
+    }
+</script>
+
+<button forward:on:click={handleClick}>
+    <slot />
+</button>
+```
+
+This pattern could be similarly applied to actions:
+
+```html
+
+<script>
+    function internalAction(node) {
+        // do something with `node`
+    }
+</script>
+
+<button forward:use={internalAction}></button>
+```
+
+In this example, the `use` directive would be both forwarded to parent context as well as `internalAction` being applied. A possible pitfall would be how the animation API would play into this, as it's unclear how the compiler would deal with multiple animations.
+
+## Potential Changes with Svelte 4
+
+It has been further proposed to completely replace the typical event forwarding syntax in the past for Svelte 4, this additional proposal would also cover that:
+
+```html
+<button on:click></button>
+```
+
+would be changed to
+
+```html
+<button forward:on:click></button>
+```
+
+This is a breaking change, and likely out-of-scope.
--- a/text/0000-forward-directive.md
+++ b/text/0000-forward-directive.md
@@ -27,7 +27,7 @@ At the moment, there are a long list of pitfalls with community-provided compone
     on:keypress
     on:keydown
     on:keyup
-    class="button {customClass || ''}"
+    class="button {customClass ?? ''}"
     {...$$restProps}
 >
     <slot />
@@ -64,7 +64,7 @@ Similar problems arise with the `transition` and `animation` directives.
 This example uses the same `<Button />` component described earlier.
 
 ```html
-<button forward:on class="button {customClass || ''}" {...$$restProps}>
+<button forward:on class="button {customClass ?? ''}" {...$$restProps}>
   <slot />
 </button>
 ```
--- a/text/0000-forward-directive.md
+++ b/text/0000-forward-directive.md
@@ -148,7 +148,7 @@ This pattern could be similarly applied to actions:
     }
 </script>
 
-<button forward:use={internalAction}></button>
+<button forward:use:internalAction></button>
 ```
 
 In this example, the `use` directive would be both forwarded to parent context as well as `internalAction` being applied. A possible pitfall would be how the animation API would play into this, as it's unclear how the compiler would deal with multiple animations.


Tropix126:
Have updated with thoughts from my comment above.

kevmodrome:
Just re-read the part on `Additional Proposal - foward effects and singular event forwarding`. This is trivial to get around at the moment for actions: `<button forward:use use:action />`. The same can be done with events: `<button forward:on on:click={internalFunction} />`. 

Where it might make sense is in the following situation where you want to run a function on all events: `<button forward:on={internalFunction} />` vs `<button forward forward:on on:*={internalFunction}`.

kevmodrome:
> I really like forwarding specific events and actions (maybe transitions and animations too?) with `forward:on:event` & `forward:use:action`. Should definitely make it in.

`forward:on:event` is not needed, you can already do that today by just writing `on:event` :)

PatrickG:
> I really like [...] & `forward:use:action`. Should definitely make it in.

What should `forward:use:action` do? `action` is a function that the component author specifies. It gets executed, and then?
I guess the way to go would be `forward:use use:action`.

tropicaaal:
The whole last part involving Svelte 4 was based on some discussion I saw in the `on:*` PR in which @benmccann suggested a more explicit syntax for forwarding events rather than not passing a handler. I agree that we should avoid using double directives as much as possible, and if there's a way to do that that's also easy to understand, i'm all in.

As @kaisermann said, we can also alter the side-effect syntax to use both the `on/transition/use` and `forward` directives:

```html
<button forward:on:click on:click={internalFunction}></button>

<button foward:in in:internalTransition={}></button>
```

We could also just ignore the additional proposal and only use `forward` as a method of routing the whole directive rather than a specific event type, which would avoid "double directives" completely, which is what i'm for initially.

kevmodrome:
> I don't think we need any extra syntax for exposing transitions/actions. You can expose that through props alreay, which is top-down, and introducing a `forward`-syntax for that, which would be bottom-up, is confusing to me. So the RFC would be reduced to `forward` of events, at which point the question is if `bubble` is the better wording.

I don't think this is as clear cut as you make it out to be. For transitions I see your point and I think I agree but for actions it's a bit trickier. This makes it a lot easier to apply multiple actions, rather than having to pass in an array of them into some prop. Applying multiple actions dynamically is tricky at the moment - you have to write a utility action that takes an array of actions. Compare these two as a consumer of the component:

```html
<MyInput use:autoResize={400} use:copyOnClick={{ once: true }} />
```
vs
```html
<MyInput actionsToApply=[{action: autoResize, props: 500}, {action: copyOnClick, props: { once: true }}]
```

Another thing it has going for it is that immediately obvious that there is an action or transition applied to the component, I think that's worth something.

tropicaaal:
> Regarding actions/passing in stuff: This is true for the case where you want to expose one specific element, but fails when you want to be able to apply actions/transitions/whatever to multiple DOM elements, at which point it could get more confusing. Conceptionally, an DOM element consists of - well - one element, whereas for components it could be 0-x. Regarding "leave it to authors to abstract away svelte language syntax": I don't think this is particulary painful. Writing `export let transition; ... transition:transition` is only a couple more keystrokes than `forward:transition`. I feel your desire to have a uniform API (this `actions=[{action: .., x: ..,y: ..}` doesn't feel very idiomatic), but I don't think it's that tradeoff-free as it is presented here.

This doesn't necessarily invalidate the use of `use` props, it just provides an easier pathway for the most common scenario. $$restProps is a similar example in my eyes. You *could* export a prop for every possible attribute someone might need or you could use the convenient solution and route all of the attributes to a single DOM element in your template.

Although custom action syntax might not seem like a lot, it ends up stacking up when you have to deal with *every* edge case over a large amount of components.

tropicaaal:
> This could be expanded to cover the use case for forwarding CSS variables,
> 
> ```
> 
> <button forward:--btn-color></button>
> ```
> 
> ```
> 
> <Button --btn-color="#007FFF" />
> ```
> 
> Unless I'm missing something, this works around the issue of css vars on components potentially messing up the css structure with the extra div, since we now know exactly which element to put `style="--btn-color: #007FFF;"` on. Makes themable components a lot more easier.

I've discussed this in the discord a bit, but I think handling style props is out of the scope of this RFC. Recently, the PR for the `style:` directive was approved and it'll likely be merged soon, so this proposal *can* handle forwarding style directives, however I believe the `forward:` syntax should keep the single consistent purpose of forwarding directives in components. Introducing a one-size-fits-all syntax for handling directives and style props (classes were proposed too at some point) would just make things more difficult to understand.

rgossiaux:
Personally I really like this direction and think that something like this is sorely needed in Svelte. 

What are the semantics around event dispatching? eg if I have something like
```
// MyComponent.svelte
...
dispatch("update", value);
...
<ChildComponent forward:on />
```

If I do `<MyComponent on:update={...} />`, are we listening to an `update` dispatched from `MyComponent`? From `ChildComponent`? Both?

Florian-Schoenherr:
@rgossiaux what happens if you currently do this, but forward only update?
edit:
https://svelte.dev/repl/50b98a9d3d2a46e7b2ba1c1eff063130?version=3.44.3
so it sends both (and keeps the order, lowest first)

tropicaaal:
> Personally I really like this direction and think that something like this is sorely needed in Svelte.
> 
> What are the semantics around event dispatching? eg if I have something like
> 
> ```
> // MyComponent.svelte
> ...
> dispatch("update", value);
> ...
> <ChildComponent forward:on />
> ```
> 
> If I do `<MyComponent on:update={...} />`, are we listening to an `update` dispatched from `MyComponent`? From `ChildComponent`? Both?

Did a little testing over how Svelte handles this currently with conventional forwarding, and it's both:
https://svelte.dev/repl/6984269f1c9d428b91412ef93d063195

rgossiaux:
Worth mentioning it in the RFC if that's how `forward:` should behave too. I called this out because I think that won't necessarily be desired all the time with custom events. In my example, the author of this component certainly intended to expose the `update` event from the outermost component, and they may wish to use `forward:` to forward all DOM event handlers to a lower component, but they might not have intended to expose any internal `update` events further down the tree. Any component that both dispatches events and uses <svelte:self> or otherwise includes itself may run into this, for example, in a way that might be annoying to resolve.

Some possibilities for dealing with this case:
* Ignore it for now. It won't be a problem most of the time, especially if your component doesn't make much use of custom events.
* Provide some way to only forward DOM event listeners. Then custom events could be forwarded individually with `on:`. The set of custom events for a component will be known and finite (and presumably small). But this is also a bit gnarly since now the framework is making some decisions about if your event is a DOM one or not.
* Provide some way to exclude individual event names from `forward:on`. I kind of like this option the most as long as it's not too unwieldy.
* Do forwarding in a different way from `forward:on` entirely. For example, an alternate way would be for the event handlers to be available to the component in some object which you could spread onto a child component, which would give you full control over which handlers to forward or not forward.

Florian-Schoenherr:
> to forward all DOM event handlers to a lower component

isn't forward meant the other way around? (forward the event up the tree)

anyways, excluding seems like it would be useful, say you want a component to handle click in some way and dispatch it explicitly afterwards (or with changed event details).

I thought about syntax, which is a little unwieldy, though that might be _okay_ as you would mostly only exclude some events:
```svelte
<button forward:on exclude:on:click/>
<!-- or, to serve my example above: -->
<button forward:on exclude:on:click={(e) => dispatch("click", doSomething(e))}/>
```

tropicaaal:
> > to forward all DOM event handlers to a lower component
> 
> isn't forward meant the other way around? (forward the event up the tree)
> 
> anyways, excluding seems like it would be useful, say you want a component to handle click in some way and dispatch it explicitly afterwards (or with changed event details).
> 
> I thought about syntax, which is a little unwieldy, though that might be _okay_ as you would mostly only exclude some events:
> 
> ```svelte
> <button forward:on exclude:on:click/>
> 
> <button forward:on exclude:on:click={(e) => dispatch("click", doSomething(e))}/>
> ```

There's a slight issue with this in particular, notably the fact that you are using this exclude directive on a specific event instance, not an event name. I think that `forward:on={{ exclude: ["click"] }}` might make more sense in the context of an API. If we do decide to exclude event instances, `on:click|exclude` might work too and not require the addition of a new directive. Would be interested to hear other people's thought's on this.

Florian-Schoenherr:
just realized `exclude:click` would be shorter.
I don't really like strings there (intellisense), second example is good.

iacore:
This feature implies that the Svelte compiler should know all the event types of any element, and the namespace of element used. Is `button` in HTML deserve this feature more than `circle` in SVG?

tropicaaal:
Not sure if this needs an explicit update to the RFC, but this could also fit in with the `style:` directive now that it's merged.

tropicaaal:
> This feature implies that the Svelte compiler should know all the event types of any element, and the namespace of element used. Is `button` in HTML deserve this feature more than `circle` in SVG?

Would it? I don't think that the compiler would care about what events are passed down, it would just add the respective listener to the item with the forward directive. For reference, forwarding all events is actually already possible by [monkeypatching `SvelteComponent.$on` and abusing internals](https://github.com/Tropix126/fluent-svelte/blob/main/src/lib/internal.ts#L119), but I don't think I need to explain why that's less than ideal.

iacore:
What I said was related to DOM API `addEventListener`, where the list of events that might generate from a DOM element will change over time due to change in W3C standards.

tropicaaal:
Hmm, so I've been trying to think of how event exclusions should work. The only way to really do this is to exclude via an event name. Something like `on:click|exclude` wouldn't work because that's attached to the specific event's instance rather than a given name (i.e. "click").

There are a few possible approaches to this sort of syntax:

## 1. Some sort of config object built-in.
```html
<div forward:on={{ exclude: ["click", "somedispatchedevent"] }}>
```
### Pros:
- Allows for exclusions without any extensive additional syntax.
- Opens the possibility to excluding events based on things like regex - it's just a javascript string.

### Cons:
- The problem is that this sort of API can lead to a bit of jank, as no other directives other than actions allow for this sort of configuration.
- Some questions arise on how much complexity this actually brings. I'd imagine that allowing javascript string methods and whatnot in this sort of config setup would result in some weird edge-cases.

## 2. Add a dedicated exclude directive.
```html
<div forward:on exclude:click exclude:somedispatchedevent>
```
### Pros:
- The syntax is concise and there isn't really anything special going on.

### Cons:
- Requires the addition of another directive.
- Might be a pain to use if you have many of them chained together.

## Some thoughts

I've also considered just not dealing with event exclusions. After all, the original intention of the `forward:` directive was to allow you to delegate directives to a component declaration, and event exclusions could be seen as a sort of abstraction leak from this point of view.

The main question here is how much of a limiting factor would not considering this be? Obviously, if you were writing an internal library your users would have access to internally dispatched events, although they would probably have to go out of their way to find this.

The current "solution" pioneered by SMUI and now used by headlessui-svelte and my fluent-svelte project is to monkeypatch `SvelteComponent`'s `$on` property. That looks something like this (with varying syntax differences):

```html
<script>
    import { createEventForwarder } from "$lib/internal";

    const forwardEvents = createEventForwarder(["excluded", "excluded2"]);
</script>

<div use:forwardEvents></div>
```

This has a few problems, including but not limited to:
- Forwarded events are recognized by language tools as `CustomEvent`s, which has some problems when working with typescript.
- Any dispatched events through `createEventDispatcher` *must* be specified for exclusion else they won't fire, but this proposal's system will both forward and fire dispatched events.
- Abuse of internals, etc etc.

Thoughts? @rgossiaux

rgossiaux:
@Tropix126 your most recent comment makes sense to me (though just to note that I added the event exclusion stuff to the code in svelte-headlessui; it's not there in SMUI).

After letting this simmer for the last month or so, I increasingly think the event exclusion thing is important. For one thing, I ran into it *right away* when trying to use this kind of forwarding. Beyond that, though, it also lets you do stuff like:

```svelte
<button on:click>Press me</button>
<ChildComponent forward:on={{ except: ["click"] }} />
```

Re: syntax for this, I think what I just posted above is also my my preferred option at the moment. I don't like `<Component forward:on except:click />` because the `except:click` has a dependence on `forward:on`. And I think `except` sounds a little more natural to me than `exclude`: *forward* the *on* directive, *except* for *click*.

I wouldn't want this to block this feature from getting merged but I strongly suspect that if it gets merged without it, people are going to realize that they want/need the ability to exclude certain events once they start using it, and we'll have to go through this whole merry-go-round once again. Hopefully we can just get it right from the start.

I think it's pretty natural, for what it's worth. Right now your options are "forward no events" and "forward no events, except for this finite set". We'd be adding "forward all events" and I think "forward all events, except for this finite set" is a very natural counterpart.

I also have grown to appreciate React's approach, to be honest. Simply passing down the handlers explicitly has some advantages--components are now free to forward them in whatever way is appropriate, including in more exotic ways that this directive still won't solve (for example, in `svelte-headlessui`, I need to pass around a [click handler](https://github.com/rgossiaux/svelte-headlessui/blob/cdc0d860447f415b818262898f2424f9c2ed7481/src/lib/components/switch/SwitchGroup.svelte#L48) via context in some cases so that clicking on a label will take action on its corresponding form control). Of course there's a huge downside with the React way too--multiple listeners for the same event will clobber each other. But I still find it useful to keep in mind.

Florian-Schoenherr:
Why does everyone want more "{{[]}}" and less types for events? 😄 
I still think `<Component forward:on except:click />` is quite simple, because the dependence doesn't really add much work? Just needs a compiler error if you do `except` without forward.
If it's easier to check the string inside the `{{except:[".."]}}`, I'm with you

tropicaaal:
I will say that I personally believe `except` to be a bit generic of a name for it's intended purpose, and might be confusing for newcomers to comprehend what it does.

aradalvand:
I also like the `<ChildComponent forward:on={{ except: ["click"] }} />` approach better.

lukaszpolowczyk:
Proposal ORDERED `except`, etc.:
Named forward groups.

In a proposal, you can pass any special attribute to any element inside a component.

But what if you make named forward groups?
For example:
```svelte
<Button
  group-name1::on:click={_=>console.log("group-name1 click")}
  group-name1::on:mouseup={_=>console.log("group-name1 mouseup")}
  group-name2::on:click={_=>console.log("group-name2 click")}
  group-name2::on:mouseup={_=>console.log("group-name2 mouseup")}
>I'm a button with an action!</Button>
```
```svelte

<button forward:group-name1::on>
  <slot />
</button>

<button forward:group-name2::on>
  <slot />
</button>
```

This would have the result:
```svelte

<button 
  on:click={_=>console.log("group-name1 click")}
  on:mouseup={_=>console.log("group-name1 mouseup")}
>
  <slot />
</button>

<button 
  on:click={_=>console.log("group-name2 click")}
  on:mouseup={_=>console.log("group-name2 mouseup")}
>
  <slot />
</button>
```

The `::` character seems sensible, since it is not reserved and is used in some languages for namespaces.

That would solve one problem with the forward directive proposal.
What do you guys think?

---

EDIT: O! It could also be suggested to add multiple groups to one special attribute, e.g.:

```svelte
<Button
  group-name1::on:click={_=>console.log("group-name1 click")}
  group-name1::group-name2::on:mouseup={_=>console.log("mouseup")}
  group-name2::on:click={_=>console.log("group-name2 click")}
>I'm a button with an action!</Button>
```
No change in the component:
```svelte

<button forward:group-name1::on>
  <slot />
</button>

<button forward:group-name2::on>
  <slot />
</button>
```

This would have the result:
```svelte

<button 
  on:click={_=>console.log("group-name1 click")}
  on:mouseup={_=>console.log("mouseup")}
>
  <slot />
</button>

<button 
  on:mouseup={_=>console.log("mouseup")}
  on:click={_=>console.log("group-name2 click")}
>
  <slot />
</button>
```

Say, doesn't this solve all the problems with "except"/"exclude" etc. etc.? :)

brandonmcconnell:
> Why does everyone want more "{{[]}}" and less types for events? 😄 
> 
> I still think `<Component forward:on except:click />` is quite simple, because the dependence doesn't really add much work? Just needs a compiler error if you do `except` without forward.

I also prefer that `except:click` syntax because it's simpler, but I think it would need to be more specific, but not quite as complicated as the suggested `{{[]}}` syntax.

Imagine an element which needs to forward multiple directives, and exclude (except) for one of them. The entire directive should be used, like this:

**Example #1**

```svelte
<div
  forward:on
  except:on:click
  except:on:mouseup
  forward:transition
  except:transition:fade
/>
```

Or more consolidated (no preference here):

**Example #2**

```svelte
<div
  forward:click|transition
  except:on:click|on:mouseup|transition:fade
/>
```

Or to support multiple handlers per event, you could use the pipe modifiers once per line without repeating the event, like this:

**Example #3**

```svelte
<div
  forward:click
  except:on:click|mouseup
  forward:transition
  except:transition:fade
/>
```

I personally prefer either the first or third example ☝🏼, and the third is my personal favorite. It would still support the first example's syntax if someone wanted to split the `except` further, while offering the flexibility of some consolidation, similar to the second example but without inflating each directive with so many explicit calls (having to use the event name with each instance).

Forwarding a simple event handler like `on:click` would be supported as well, though this is already natively supported by simply using `on:click` without `forward:`.

```svelte
<div forward:on:click />
```

would render the same result as the below example in Svelte v3:

```svelte
<div on:click />
```

lukaszpolowczyk:
@brandonmcconnell I feel transparent like a ghost, as if I have shadow ban.

brandonmcconnell:
@lukaszpolowczyk Push through. It's easy for PRs and issues ti get lost along the way. If you want to draw more attention to this, try tweeting about it and tagging both [@sveltejs](https://twitter.com/sveltejs) and [@SvelteSociety](https://twitter.com/sveltesociety). I would also suggest posting about your PR in Svelte's official Discord channel, which you can access via [svelte.dev/chat](https://svelte.dev/chat).

Best of luck! I really like this proposal, as I think the ability to forward any directives would greatly improve Svelte, especially in the case of transitions (in my case).

MentalGear:
Great proposal, especially for using style: on custom components!

aradalvand:
I think this is almost a must-have for Svelte, in any serious development you'd come across cases where a feature like this would be immensely useful.

emmnull:
Found this rfc after having to jerry-rig things with `create_[direction]_transition()` functions and exported props to make components transitionable without the need for additionnal wrapping elements. This really feels like something interesting for a clean DX when working with primitive/atomic components.

tropicaaal:
It's been a bit since I last touched this conversation, although these are my personal thoughts on some matters now.
- I'm still generally against a second directive (like `except` or `exclude`) due to some ambiguity in it's control flow. For example, what happens when it's used alone or used with traditional forwarding of singular events? Additionally it's not really a pattern observed before in Svelte where two directives "communicate" in this way.
- I'm also against introducing new syntax patterns for directives in general just to accomodate for an edge case. HTMLX is a standardized specification that svelte adheres to and its more likely that this RFC will be considered if we don't go that route. This includes "double directives" (`x:y:z`) or other new changes to template syntax that aren't really used anywhere else.
- The more I think about exclusions the less I want to think about them. I don't know if there's a good non-leaky way to introduce them (keep in mind forward isn't just targeting events, and a base level config for excluding them could be ambigious) and it might be fair to just not address it if an initial implementation is made. If this RFC does get merged then exclusions could be considered after the fact since any new syntax here likely wouldnt be breaking.

lukaszpolowczyk:
@Tripix126 What about my proposal for naming and grouping? - https://github.com/sveltejs/rfcs/pull/60#issuecomment-1073263370

It doesn't go beyond HTMLX. There is only a double colon `group::`.

tropicaaal:
> @Tripix126 What about my proposal for naming and grouping? - [#60 (comment)](https://github.com/sveltejs/rfcs/pull/60#issuecomment-1073263370)
> 
> It doesn't go beyond HTMLX. There is only a double colon `group::`.

I think it *could* fit into what the forward direct offers, but is definetly out of scope for this initial RFC. If something like this were to be added id expect it to come in a new major version where syntax or control is being rethinked. I generally want to keep the scope of this RFC to be as narrow as possible, and not try to solve everything (i.e. trying to fix style scoping or introducing new conventions in even handling).

lukaszpolowczyk:
@Tropix126 Whereas I would like it to work ONLY with named groups.
That would make the action of `forward:` absolutely unambiguous, even if someone doesn't know the syntax. The group name itself makes you know that what you added in the `group::on:click` component call is the same as what you gave in, for example, `forward:group::on:click`.

Of course, it doesn't have to be a double colon, it's just a suggestion, but just naming the group makes the syntax much more unambiguous at a glance.

And the fact that it provides more opportunities is just another plus.

This is how I see it.

ptrxyz:
I also wouldn't like the group syntax as it looks pretty alien to me.

How about sticking to a similar syntax that we already have:

If `<button on:click>` (on:click without a handler) emits a click event to the parent component, why couldn't we have `<button forward:on>` to forward all events without an explicitly added handler the parent? 

So with `forward:on` all events without a handler are emitted. If you want to handle it inside the child component add a handler. If you then also want to forward the event, you can manually add `dispatch` to your handler to make it bubble to the parent.
If you do not want to emit an event, how about adding a modifier such as `|stopPropagation` or something similar.

(Analog to `|stopPropagation`, one could also argue to add a explicit modifier `|doPropagate` for the events that I have a handler and that should also be forwarded to the parent. I don't think that's how Svelte does it for now, so it would probably be a bit of an extra thing. )

```jsx
<button on:click>
<button forward:on> 
<button forward:on on:click on:focus={something}>
<button forward:on on:focus|stopPropagation>
```

Example 1: forward click events only
Example 2: forward all events
Example 3: forward all events (incl. click), but not focus, unless the function tells you to do so.
Example 4: forward all events except focus

For example 3 one could argue, that focus should be forwarded, just as the other events, unless `|stopPropagation` is added. However i think as soon as you attach a handler explicitly, you can also dispatch the event manually if you want it to bubble up. 

And of course it doesn't need to be `stopPropagation`, if people think this is too loaded then. How about `noForward` then?

tropicaaal:
@ptrxyz Your fourth example would still forward `focus`.
This is *interesting*, but how would it consistently tie into forwarding other directives?

wickning1:
I just came across this today and realized it seeks to solve essentially the same problem we’re discussing in https://github.com/sveltejs/svelte/issues/5218

ptrxyz:
> @ptrxyz Your fourth example would still forward `focus`. This is _interesting_, but how would it consistently tie into forwarding other directives?

Example 4 was meant to show how it would be possible to exclude events from being forwarded by adding a directive like `|stopPropagation`. I chose this, was it fits quite nicely in my mind. It stops the event from bubbling.
I see this from the "child to parent" direction: the child component defines what should be forwarded to the parent, in example 4 this would be every event, however `focus` is stopped from propagating upwards, so it would not reach the parent.

For me, `forward` should not try to be ManBearPig and try to solve everything. It's a way to cope with a simple problem: what's the "catch all" event handler and how to expose those events to the outside world, so we don't have to do `on:X`, `on:Y`, `on:Z` etc..

For anything else, I would probably introduce a new keyword as it's a totally different cup of tea. Let me explain my thought there:
For me `actions` and `transitions` are a different from the 'catch-all-event'-thing since the direction is the other way around: the child should 'attach'/use the actions passed down from the parent and not hand them upwards.
Further the parent is a single node, the child just hands over a bunch of events and the parent knows how to deal with it. On the other hand, the child consists of 0..many nodes and we seem to want to pass down 0..many actions and transitions. This has implications on the syntax, as the child needs a way to address the passed down things and associate them with the correct nodes it consists of.

ptrxyz:
Maybe an idea on how to solve the 2nd part. Look at this code:


```svelte
<!-- Parent -->
<MyComponent
    on:focus={willNeverHappen}
    on:keyDown={handledInParent}
    on:click={handledInParent}
    use:actionForChild={action1_params, identifier="action1"}
    use:actionForChild={action2_params, identifier="action2"}
    transition:fade={animation1_params, identifier="anim1"}
    transition:slide={animation2_params, identifier="anim2"}
>

<!-- this is consistent with the syntax we have now for DOM nodes: -->
<button
    on:click={handledInParent}
    transition:fade={animation1_params, identifier="ThisWillNeverBeUsed"
>
```
Here, the identifier is an optional property to keep things consistent with the "normal" signature for transitions/actions but makes it possible for the child to identify them.

The child component can now access the `action`/`transition` handlers by their ID:

```svelte
<script lang="js">
import { defaultAction } from "$actions/my_default_action'

function FunctionReturningAnAction(node) {
    return $$attachables?.use?.action1 || someDefaultAction || undefined
}
</script>

<!-- Child Component -->
<button
    forward:on
    attach:use={FunctionReturningAnAction}
    attach:transition={(this) => $$attachables.use.transition1 || undefined}
    on:click
    on:focus|stopPropagation
    on:keyup={handleInClient}
>

<button
    forward:on
    attach:transition={"transition2"}  // <-- should this be allowed?
    on:click
    on:focus|stopPropagation
    on:keyup={handleInClient}
>
```
I'd introduce a directive called `attach` that attaches a `thing` that's identified by the part after the colon `:`:

`attach:use` --> attach an action
`attach:transition` --> attach a transition
`attach:whatevermightcome` --> for future use

The directive expects a function that returns a matching object, so for `attach:use`, the function should return an `action` object for `attach:transition`, the function should return a `transition` object etc. If we want it fancy, one could make a case that it would be enough to simply put a string there which is then "auto-resolved" to the matching object from the $$attachables object, however for the sake of general usability, I wouldn't see that as a good idea.
To access the "attachables", one could introduce something similar to `$$props`, let's call it `$$attachables`, which is an object that holds all the attachables from the parent. In my example the child would receive an object like this:

```js
$$attachables = {
    "use": {
        "action1": <callable here that represents the action `actionForChild({action1_params})`>,
        "action2": <callable here that represents the action `actionForChild({action2_params})`>
    },
    "transition": {
        "transition1" : <callable here that represents the transition `fade({animation1_params})`>,
        "transition2" : <callable here that represents the transition `slide({animation2_params})`>
    }
}
 ```

So tl;dr:
- Syntax to pass down stuff to children: same as for DOM nodes, but with optional 'identifier'; instead of only allowing a single 'use:' and 'transition:' directive, we make sure the tuples `<directive, id>` are unique.
- Passing down attachables: "attachables" are made available to the child in a `$$attachable` object
- Syntax to access attachables in the child component: consistent to `use:<action>={{params}}` or `transition:<transition>={{params}}`, in this form `attach:use=<function that returns "action({params})">` respectively `attach:transition=<function that returns "transition({params})">` (==> general notation `attach:<something>=<function that returns "something({params})">`).

**Comment by lukaszpolowczyk** [comment]
> I also wouldn't like the group syntax as it looks pretty alien to me.

@ptrxyz If this seems extraneous, I'll give another idea.
Also, someone may not like it, but it's hard, everyone won't get along:

### Targeted Slots
Description and example - https://github.com/sveltejs/rfcs/pull/41#issuecomment-1126434719

This is a suggestion for a proposal, on [Declarative Actions](https://github.com/sveltejs/rfcs/pull/41#issue-751053559).
The author of the proposal Declarative Actions liked it - https://github.com/sveltejs/rfcs/pull/41#issuecomment-1172782667

I think Targeted Slots solves the same problems. Although the original targets were not identical.

This would solve the problems of **Forward Directive** and **Declarative Actions** all at once.

**My example of "[named forward groups](https://github.com/sveltejs/rfcs/pull/60#issuecomment-1073263370)" turned into Targeted Slots:**

```svelte
<Button>

  <svelte:element slot=name1
    on:click={_=>console.log("group-name1 click")}
    on:mouseup={_=>console.log("group-name1 mouseup")}
  >I'm a button with an action!</svelte:element>
  
  <svelte:element slot=name2
    on:click={_=>console.log("group-name2 click")}
    on:mouseup={_=>console.log("group-name2 mouseup")}
  >I'm a button with an action!</svelte:element>
  
</Button>
```
```svelte
<!-- Button.svelte -->
<slot targeted:name1 this=button />

<slot targeted:name2 this=button />
```

This would have the result:
```svelte

<button 
  on:click={_=>console.log("group-name1 click")}
  on:mouseup={_=>console.log("group-name1 mouseup")}
>I'm a button with an action!</button>

<button 
  on:click={_=>console.log("group-name2 click")}
  on:mouseup={_=>console.log("group-name2 mouseup")}
>I'm a button with an action!</button>
```

But it's also worth seeing an example from here - https://github.com/sveltejs/rfcs/pull/41#issuecomment-1126434719

The point is that you can add tagName, and special and normal attributes inside the component if you want, and you can add when calling the component if you want.


I also call out the author of Forward Directive @Tropix126 - please evaluate.

**DISCLAIMER:** If one believes that "slots should be left alone", one could call it different. But it would literally work as upgraded slots.

rgossiaux:
> > @ptrxyz Your fourth example would still forward `focus`. This is _interesting_, but how would it consistently tie into forwarding other directives?
> 
> Example 4 was meant to show how it would be possible to exclude events from being forwarded by adding a directive like `|stopPropagation`. I chose this, was it fits quite nicely in my mind. It stops the event from bubbling. I see this from the "child to parent" direction: the child component defines what should be forwarded to the parent, in example 4 this would be every event, however `focus` is stopped from propagating upwards, so it would not reach the parent.
> 
> For me, `forward` should not try to be ManBearPig and try to solve everything. It's a way to cope with a simple problem: what's the "catch all" event handler and how to expose those events to the outside world, so we don't have to do `on:X`, `on:Y`, `on:Z` etc..
> 
> For anything else, I would probably introduce a new keyword as it's a totally different cup of tea. Let me explain my thought there: For me `actions` and `transitions` are a different from the 'catch-all-event'-thing since the direction is the other way around: the child should 'attach'/use the actions passed down from the parent and not hand them upwards. Further the parent is a single node, the child just hands over a bunch of events and the parent knows how to deal with it. On the other hand, the child consists of 0..many nodes and we seem to want to pass down 0..many actions and transitions. This has implications on the syntax, as the child needs a way to address the passed down things and associate them with the correct nodes it consists of.

I think this is a conceptual misunderstanding. The direction of all directives is parent-to-child, not the other way around. The `on:` directive passes an *event handler* down to a child and is not any different conceptually from actions or transitions.

The problem that the forward directive is intended to (partially) solve is that Svelte has a bunch of syntactical magic in the form of directives that cannot be easily interacted with programatically. In a framework like React, everything is a prop & so the programmer has total control--and wrapping components is trivial. In Svelte, however, the directives are hidden from component authors, and a side effect is that non-trivial composition of components & elements becomes unreasonably difficult. 

The idea here should be to create something which is flexible & future-proof enough to solve forwarding for both existing and future directives, but no more complicated than necessary. I also agree with @Tropix126 that getting *something* is better than nothing and if that means sacrificing exclusions then so be it, though I still believe it's a problem worth taking seriously.

lukaszpolowczyk:
@rgossiaux With my proposal, there is no problem with exclusions.
See - https://github.com/sveltejs/rfcs/pull/60#issuecomment-1204651747

Almost no new syntax. It's "just" a slight improvement of the slots.

ptrxyz:
> @ptrxyz If this seems extraneous, I'll give another idea.
> Also, someone may not like it, but it's hard, everyone won't get along:

I guess it's less about liking something, more about how consistent it is within the rest of the framework. I am not aware of any other place where a `group::`-like notation is used so far, so I'd move away from it since it seems to be a construct that would only be invented to make a single feature work. I'd rather like to see a solution that feels 'svelte' and is more in line with the rest of the framework.

About targeted slots: you suggest to remove the requirement for `this` on `svelte:element` and instead take `this` from the slot the element is slotted into? Would `<slot>` then also allow for other attributes and directives such as `class:`, `use:`, `bind:this`, `on:` and `transition:` etc? So basically the resulting node would be what you get when you create an instance of `this` taken from `slot`, attach all the attributes and directives of `<slot>` and then add (& overwrite) the attributes and directives of `<svelte:element ...>`? I assume this would only work with HTML elements and `this` can not be a Svelte component, right? 

What I mean is, if this: 
```svelte
<svelte:element slot="slot1"
    data-something="yes"
    on:click
    use:action
    class:bold={bold_condition}
>inner stuff</svelte:element>
```
plus:
```svelte
<slot name="slot1" this="button"
    data-else="no"
    on:focus={function_defined_in_child}
    class:bold={bold_condition}
    use:action2>
    <span>This is rendered when slot1 is not provided</span>
</slot>
```
would result in this:
```svelte
<button
    data-else="no"
    on:focus={function_defined_in_child}
    class:bold={bold_condition}
    data-something="yes"
    on:click
    use:action
    class:bold={bold_condition}
>inner stuff</button>
```

If I got this right, I assume it would work. I like the idea! It's also more widely usable, not just in the context discussed here.

Since we only need to transport two pieces of information then (the "what to merge" and "with what to merge this on"),  one could make it a bit more crips maybe and not need the `slot` stuff while staying true to svelty syntax:

Something like this:
```svelte
<!-- parent -->
<MyComponent>
    <svelte:element on:focus={console.log} update="native_button">inner stuff</svelte:element>
    <svelte:element on:click={console.log} update="custom_component">more stuff</svelte:element>
</MyComponent>
```
```svelte
<!-- child component -->
<button svelte:name="native_button">
<CustomComponent svelte:name="custom_component">
```
would get you this:
```svelte
<!-- renders to this -->
<button
    svelte:name='native'
    on:focus={console.log}
>inner stuff</button>

<CustomComponent
    svelte:name='custom'
    on:click={console.log}
>more stuff</CustomComponent>
```
`update` (or `merge_with` or `target` or a more suitable attribute name) can be used to specify what to merge "this element" with, `svelte:name` is used to internally name the component.


note: I was thinking if it would work to replace `svelte:name` with a simple `bind:this={some_var}` and make the component `export` `some_var`, so that the parent then could simply do `<svelte:element update={mycomponent.some_var}>`, but I assume that's probably not possible to do at compile time... ?

ptrxyz:
> ```svelte
> <svelte:element slot=name1>...</svelte:element>
> <slot targeted:name1 this=button />
> ```

Little note: Shouldn't the name be strings? `targeted="name1"` and `<svelte:element slot="name1">`

And another note: does it need to be called "targeted:name1". Can't it simply follow the syntax we already have and add a `this`: `<slot name="somename" this="button">`
Because basically in both cases you transport the same information of "this slot is named 'name1'". However in the other case, you add the extra information of "this slot is a target for something" which I am not sure if it's necessary. The `slot` doesn't need to "know" and the compiler can easily find that out by checking if there is a `this`.

So to make some rules for this approach:
- If `slot` has a `this`, we consider it a targeted slot, this means `<slot this='...'>` should create a new node of type `this` and attach all attributes & directives of `slot`, as well as every `<svelte:element>` that would be slotted into that `slot`. This is effectively what `<svelte:element>` does but extended by the merging stuff.
- If slot doesn't have a `this`, it's a normal slot and it should not accept any directives & attributes other than name. In addition it would be replaced by whatever is specified using `slot="something"` on the parent, if nothing is given default to what's enclosed by `<slot>....</slot>`

Seems a bit messy tbh. I like the shorter notation I suggested in my previous comment a bit more then. The rule set would be:
- `<svelte:element>` can be used as a vehicle to transport directives & attributes which are applied to whatever is referenced by the `update` attribute.

lukaszpolowczyk:
> I guess it's less about liking something(...).

Ok.

> About targeted slots: you suggest to(...).

Yes.


> I assume this would only work with HTML elements and this can not be a Svelte component, right?

To meet the needs of Declarative Actions and Forward Directive, it is enough to work only with HTML elements.
But the development path is open.


> If I got this right, I assume it would work. I like the idea!

I'm glad. :)

> ```svelte
> 
> <button svelte:name="native_button">.
> <CustomComponent svelte:name="custom_component">.
> ```

Here is the problem.
The `button` element.
Sometimes someone wants to give a tagName in `svelte:element` (in parent) via `this` attribute, not in child.

So either `slot` or another special tag must be left.

In addition, the slot allows you to pass a property from child to parent, via `let:val` - that would be missing. I described in an UPDATE here - https://github.com/sveltejs/rfcs/pull/41#issuecomment-1126434719
Worth reading.
I want Targeted Slot to have all (almost all?) slot functions + more(what is in Targeted Slot).

It doesn't have to be called `slot` etc, names can be found better.

> note: I was thinking if it would work to replace svelte:name with a simple bind:this={some_var}. 

This is rather out of the question. See Declarative Actions - there the most important thing is that `bind:this` can be used in the normal way.

Also consider that the normal `<slot>` can be used multiple times. I'd like to keep this too, as it also gives more possibilities for Targeted Slot.

---

> Little note: Shouldn't the name be strings? targeted="name1" and <svelte:element slot="name1">.

This is a string. According to the syntax, the absence of quotation marks is still a string. Quotation marks are only necessary if the string has whitespace characters. But this is a matter of pure HTML.

> And another note: does it need to be called "targeted:name1". Can't it simply follow the syntax we already have and add a this: <slot name="somename" this="button">.

As I wrote before, sometimes you want to add a tagName in `<svelte:element this="div">`. Then it wouldn't be possible.
Another thing is that slots give you the possibility to use multiple slots, with the same name (slots allow this - I've already written too), in the same child component. Then it would also not be possible.

So there must be `targeted:name` or something similar.

The third thing is that the `<slot>` passes parameters, and you have to pass them somehow in Targeted Slots. I also wrote about this in the UPDATE.


**Summary:**
* Attribute/tag names can be changed if you find a better one. And not to confuse regular slots, with Targeted Slots.
* I do not exclude the use of components as Targeted Slot.
* I want Targeted Slots to take everything they can from regular slots.
* I am very happy that already the second person liked it. I wonder what the people who develop the SvelteJS engine think about this and know how to implement it technically.

If @Tropix126 also says it's a good solution, and someone from the SvelteJS maintainers says it makes sense, then rfcs should be done.
Because for now I described it in the Forward Directive thread and in the Declarative Actions thread. It would be better to have it in one place.

ptrxyz:
So, can you summarize how this would blend in with the topic of this RFC? I get how "beefing up" slots is a thing, but I wonder if it's also the best solution for the problems presented here.

As an example, how would it look when you have a third party component library and you want to capture focus, click and keyup events from a parent component. As of now, one would have to verbosely write down all the event listeners in the child component.

Could you for demonstration purposes rewrite this if your idea was implemented?

```svelte
<script>
    // simple child component
    export let caption = "Button";
</script>

<button on:click on:focus on:keyup>{caption}</button>
```
```svelte
<script>
    // parent component
   import Button from './button.svelte'
</script>

<Button on:click={console.log} caption="Hello!" />
```
REPL: https://svelte.dev/repl/eb2fb28276a442ff802fce2c9bc53fa1?version=3.49.0

lukaszpolowczyk:
@ptrxyz This way:
```svelte
<script>
    // simple child component
</script>

<slot targeted:button this="button" on:click on:focus on:keyup/>
```

```svelte
<script>
    // parent component
   import Button from './button.svelte'
</script>

<Button>
  <svelte:element slot="button" on:click={console.log}>Hello!</svelte:element>
</Button>
```

And frankly, you can consider a slot without a name when there is only one slot. Because slots allow only one slot. But I'm not sure, because then it loses the advantages of passing attributes through `svelte:element`:

```svelte
<script>
    // simple child component
</script>

<slot targeted:default this="button" on:click on:focus on:keyup/>
```

```svelte
<script>
    // parent component
   import Button from './button.svelte'
</script>

<Button on:click={console.log}>Hello!</Button>
```

This is basically almost the same as `forward:directive`. But after using named slots, it gives you more possibilities. When you have named slots, you have something like exceptions, etc. (by using the names)

Unfortunately, it requires the use of the `targeted:default` attribute. Alternatively, it would not be `<slot>`, but e.g. `<target>`, and then it could be treated the same as a slot without a name, but with extra Targeted Slots functions.


---

I wonder what to do with the "default content" slots `<slot>this default content</slot>`.
Sometimes you want to set the content inside the child component.

With Targeted Slot you can do it like this:

```svelte
<script>
    // simple child component
    const content = "Button";
</script>

<slot targeted:button={ {content} } this="button" on:click on:focus on:keyup/>
```

```svelte
<script>
    // parent component
   import Button from './button.svelte'
</script>

<Button>
  <svelte:element slot="button" on:click={console.log} let:content>{content}</svelte:element>
</Button>
```
...BUT sometimes you want to put there not only text, but more HTML sub-elements. That can't be done.
This is a thing to think about.
If you would add the possibility of more complex content, then it certainly can no longer be called `<slot>`, but in a different way. Because that would already be an alternative action than `<slot>`.

In theory, something like:

```svelte
<script>
    // simple child component
</script>

<target targeted:button this="button" on:click on:focus on:keyup>Button <span>someting</span></target>
```

```svelte
<script>
    // parent component
   import Button from './button.svelte'
</script>

<Button>
  <svelte:element slot="button" on:click={console.log}></svelte:element>
</Button>
```
...Where `Button <span>something</span>` is not instead of `svelte:element`, but inside `svelte:element`. E.g. provided that `svelte:element` has no content.
That is, adding content to `svelte:element`, would overwrite the content from inside `<target></target>`.

This could also be done in other ways, but I'll let it go for now.

lukaszpolowczyk:
@Tropix126 Check out my RFCS:

It solves all problems and more.
There are examples.


I'm sure @ptrxyz will be interested.

FeldrinH:
The proposal says:
> * There are some directives that could not be forwarded, most notably bind.

I'm not very familiar with Svelte internals so maybe I'm missing something. Why can't bind directives be forwarded?

TheOnlyTails:
`bind:` is, by definition, acting on the whole component, not just the markup. That's why forwarding it means nothing. This proposal mostly applies to directives that cannot be used on components, like `use:`, `style:`, etc.

tropicaaal:
> I'm not very familiar with Svelte internals so maybe I'm missing something. Why can't bind directives be forwarded?

First and foremost, `bind` already has predefined behavior for components. `bind:this` on a component get's you the component's constructor, and `bind:property` will initiate a two-way bind between a component property and an existing variable. Both of these conflict with the concept of what `forward` brings to the table, and most functionality offered by `bind` can already be "forwarded" through property binding (or othewise probably shouldn't be forwarded at all).

For example, you can easily simulate a "bind:this" to a component's inner element by exporting a prop like so:
```html
<script>
	export let someEl = null;
</script>

<div bind:this={someEl} />
```

then access it like

```html
<Component bind:someEl={someVariable} />
```

A similar story applies to `bind:group` and `bind:value`, and many others as well. Just export a corresponding `group` or `value` prop and use property binding.

lukaszpolowczyk:
@Tropix126 Why are you ignoring me?
I hacked some work into my RFC and would like you to review it.

tropicaaal:
@lukaszpolowczyk I've been rather busy as of recent and haven't had time to look through it, although i've seen your comments. I'll have a look when I get the chance.

adiguba:
Hello,

I haven't see this before sending an similar issue.

Just one remark : I think that filtering the forward type is ambigous. If a component forward to a node, all directive should be usable...

PR #61, opened 2021-10-12
Remove `{#await}`, add a `promisable` store

Prinzhorn:
[Rendered](https://github.com/Prinzhorn/svelte-rfcs/blob/remove-await/0000-remove-await.md)


Prinzhorn:
@lukeed you are right, I didn't look at it that way

> If you're saying that `promisable` can _only_ exist if `#await` is removed – which I don't think you're saying

Right, I think having `promisable` in Svelte core would be beneficial regardless of what happens with `{#await}`. So please consider it separate. But if it were to be added, than the question is what benefit keeping `{#await}` would have. I guess that depends on how well `promisable` would be received and what the implications for SSR are (that's what I'm most worried about, I'm not sure if SSR currently works properly if I don't use `{#await}`, I think it'll just send the HTML in `pending` state over, because how would it know what I'm doing).

PR #63, opened 2022-01-23
style blocks

pasqui23:
[rendered](https://github.com/sveltejs/rfcs/blob/ddb2dbba1d94e3008838c0838fd19c8d99241468/text/0000-style-blocks.md)

### DIFF
--- a/text/0000-style-blocks.md
+++ b/text/0000-style-blocks.md
@@ -1,0 +1,122 @@
+- Start Date: 2022-01-22
+- RFC PR: (leave this empty)
+- Svelte Issue: (leave this empty)
+
+# Style blocks
+## Summary
+
+A way to let components style their child
+## Motivation
+
+A way to style child from parents has been repeatedly asked and it has each time refused as it ofter amounted to simple syntactic sugar for `* :global(some-selector)` CSS selector.
+While syntactic sugar would not be bad per any solution based on the above selector would pass the style to every child,not only its immediate, without any way to reset styles.
+
+We propose a way for Parent to style only their immediate children, without breaking style encapsulation.
+## Detailed design
+```svelte
+<!-- Parent.svelte -->
+<script>
+import Child from ‘./Child.svelte’
+</script>
+<Child>
+Hello world!
+</Child>
+<style>
+Child-block1{
+  background-color:purple;
+}
+</style>
+
+<!-- Child.svelte -->
+<script>
+import Child from ‘./Child.svelte’
+</script>
+<div><slot /></div>
+<style>
+div{
+  -svelte-block:block1;
+}
+</style>
+```
+will be rewritten to
+```svelte
+<!-- Parent.svelte -->
+<script>
+import Child from ‘./Child.svelte’
+</script>
+<Child __classes={{block: `svelte-Parent-Child-hash`}}>
+Hello world!
+</Child>
+<style global>
+(div).svelte-Parent-Child-hash{
+  background-color:purple;
+}
+</style>
+
+<!-- Child.svelte -->
+<script>
+export let __classes
+</script>
+<div class={classes.block}><slot /></div>
+<style>
+div{
+  -svelte-block:block1;
+}
+</style>
+```
+
+Exact syntax is open to bikeshedding but:
+* A child can declare a point in their css where parents can put styles with the syntax `-svelte-block: blockName`
+* A parent can put select such style blocks by writing the identifier of the svelte child  (https://developer.mozilla.org/en-US/docs/Web/CSS/Universal_selectors)[including `svelte|self` for recursive calls] ),a `-` and the block name.This should be  interpreted by the PostCSS preprocessor as a single tag name and the capital initial letter(or `svelte|` prefix) should differentiate it from actual tag names.
+
+In particular to pass a block style to every child a component that call itself recursively MUST write
+```
+svelte|self-blockRecursivelyPassed{
+  -svelte-block:blockRecursivelyPassed;
+}
+```
+
+Without the above CSS a child will not pass the style block to its children and a parent can only style its immediate children.
+
+### Technical Background
+
+> There are a lot of ways Svelte is used. It's hosted on different platforms;
+> integrated with different libraries; built with different bundlers, etc. No one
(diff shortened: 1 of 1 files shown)


lukaszpolowczyk:
Take a look at my proposal:
https://github.com/sveltejs/svelte/issues/6422

The advantage is that styles are passed explicitly, by property, and used in a controlled way.

pasqui23:
From what I understand of @lukaszpolowczyk 's proposal bind:styles would be actual js objects. This would hinder server side rendering as it would ask to compute the value of js objects( for example, what should the server output on `<Component styles={fetch(...)}/>`? )

lukaszpolowczyk:
@pasqui23 It would accept only one object type - a special object that is a style reference.

But I am aware that there are problems with my proposal.

Along the way, I made a preprocessor(not published) that takes a style fragment from any `*.svelte` file - such a simpler form of passing style between components.

Rather, this is not the final version, just something to think about.

pasqui23:
https://github.com/sveltejs/svelte/issues/6422#issuecomment-892659354

> I just wrote that if I did it less dynamically, it could all be prepared on the compiler side, and now it can be done, but doing some extra runtime side operations.

The problem is that neither javascript nor typescript have any mechanism to check whether a given code block is constant (and is in general a hard problem). My proposal sidesteps almost any runtime overhead by merely translating CSS selectors.

pasqui23:
> Along the way, I made a preprocessor(not published) that takes a style fragment from any `*.svelte` file - such a simpler form of passing style between components.

Well this proposal is exactly this

lukaszpolowczyk:
@pasqui23 In my case, the experiment (working) looks like this:

I used `@supports (expose(external-second-block)) {` so that it doesn't throw an error under my `expose(external-block)` syntax.
I load the styles via the `extended:external-block="url.svelte"` attribute in the `style` tag.
See that I use `extended:self-first-block="."` because the style with `external-block` is not applied to the custom component by default.

This does not use global styles at all. The styles are just copied.

The names, the way to separate `external-block`, the way to load styles to think about, can be changed (I think it may need to be rethought and changed).

```svelte
<!-- routes/index.svelte -->
<h1>svelte-component-styles-share-preprocess</h1>
<p>Visit <a href="https://kit.svelte.dev">kit.svelte.dev</a> to read the documentation</p>
<div class="self-class">self class</div>

<div class="self-first-class">self first block</div>
<div class="external-first-class">external first block</div>
<div class="external-second-class">external second block</div>
<div class="external-3-class">external 3 block</div>
<style
   extended:self-first-block="."
   extended:external-first-block="../lib/Libon.svelte"
   extended:external-second-block="$lib/Libon.svelte"
   extended:external-3-block="../lib/Libon.svelte"
>
   @supports (expose(self-first-block)) {
      .self-first-class {
         color: green;
      }
   }
   
   .self-class {
      color: blue;
   }
</style>
```

```svelte
<!-- Libon.svelte -->
<div class="external-4-class"></div>
<style>
   @supports (expose(external-first-block)) {
      .external-first-class {
         color: green;
      }
   }
   
   @supports (expose(external-second-block)) {
      .external-second-class {
         color: green;
      }
   }
   
   @supports (expose(external-3-block)) {
      .external-3-class {
         color: green;
      }
   }
   
   .external-4-class {
      color: green;
   }
</style>
```

**The result:**
obraz

I don't know if there were any problems, because I did it in November 2021. Something wasn't refreshing while editing the file, meaning there was an inconvenience during development.
[svelte-component-styles-share-preprocess.zip](https://github.com/sveltejs/rfcs/files/7963543/svelte-component-styles-share-preprocess.zip)
(`npm install` in the root directory and in the `dev` directory and then fire `npm run dev` in the `dev` directory)

pasqui23:
The exaple you have written look like a more comlicated way of doing `@import` https://developer.mozilla.org/en-US/docs/Web/CSS/@import.

And if you let the `extended:` be a runtime string /object/whatever you still go in the problem of having to compile css at runtime, while my proposal limit css generation at compile time (except for web components)

lukaszpolowczyk:
> The exaple you have written look like a more comlicated way of doing ***@***.***`.

I don't know what `***@***.***` means.

> And if you let the `extended:` be a runtime string /object/whatever you still go in the problem of having to compile css at runtime, while my proposal limit css generation at compile time (except for web components)

Have you checked?
My suggestion is preprocessor, everything is ready after compilation.

pasqui23:
> > The exaple you have written look like a more comlicated way of doing _**@**_.***`.
> 
> I don't know what `***@***.***` means.

I meant [`@import`](https://developer.mozilla.org/en-US/docs/Web/CSS/@import). My proposal is about *parents* styling *children* components, while your proposal let *children* take styles from svelte files, as you have presented [here](https://github.com/sveltejs/rfcs/pull/63#issuecomment-1024891435).
Also I think it would be better if you would post a rfc and explain in detail your alternative, clarifying any of the less clear details

lukaszpolowczyk:
I found a much better syntax - using CSS `@layer`and working similar to `set`/`getContext`:
https://github.com/sveltejs/svelte/issues/7465

The new CSS standard helped me. :)

In the preprocessor I can use this new syntax of mine and publish already without shame. :P

pasqui23:
@lukaszpolowczyk Please open a RFCand in it put an example of how a parent should put for e.g. a ~font-size:14px` in a child

@pngwn I would like to know it this would solve your concerns on a svelte theming system.

pngwn:
No it doesn't. It also has a number of limitations. 

For example it says that this would be a way of styling _immediate_ children but components are frequently made up many child components which may have any number of child components themselves. So this proposal requires them to be passed down manually, or in complex cases will require authors to set up a complex style managament system in the component itself in order to manage all of the components being passed down.

This proposal isn't very clear and lacks any detail on what the various tradeoffs are but it doesn't look like it would allow you to pass down multiple background-colors for example. It also isn't clear how you would actually make use of the passed classes or styles in the children.

This also looks like the Svelte compiler needs to take into account both parents _and_ children when compiling which it currently isn't capable of (because it only ever looks at a single file and does not retain any context about other files).

CSS variables solve this problem far more elegantly and provide far more flexibility, I can't imagine this RFC will be accepted for that reason alone.

PR #66, opened 2022-05-04
Targeted style inheritance

10p-freddo:
[Rendered](https://github.com/ghughes/rfcs/blob/style-inheritance/text/0000-style-inheritance.md)

### DIFF
--- a/text/0000-style-inheritance.md
+++ b/text/0000-style-inheritance.md
@@ -68,8 +68,8 @@ Child button span {
 ### Implementation
 
 Components that are
-- targeted by a selector, and
-- have at least one child element with the `inherit` attribute
+- targeted by a selector, or
+- have the `inherit` attribute
 
 ...are wrapped in a [`display: contents`](https://svelte.dev/repl/ea454b5d951141ce989bf9ce46767c71?version=3.14.0) div, which is assigned `data-svelte-this={name}`, where `{name}` is the name of the component.
 
--- a/text/0000-style-inheritance.md
+++ b/text/0000-style-inheritance.md
@@ -88,7 +88,7 @@ This issue has been discussed to near-death and similar proposals have been floa
 
 | Suggestion | RFC or Issue | Feedback |
 | ---------- | ---------- | -------- |
-| Allow parent to impact child component CSS | [#13](https://github.com/sveltejs/rfcs/pull/13) | Violates the principles of CSS encapsulation and components being their own boss. Does not prevent uncontrolled cascade. |
+| Allow parent to impact child component CSS | [#22](https://github.com/sveltejs/rfcs/pull/22) | Violates the principles of CSS encapsulation and components being their own boss. Does not prevent uncontrolled cascade. |
 | Component-scoped `<style>` | [comment in #13](https://github.com/sveltejs/rfcs/pull/13#issuecomment-553144981) | Breaks encapsulation. |
 | Unscoped CSS by default | [svelte#901](https://github.com/sveltejs/svelte/issues/901) | Scoped CSS by default is an intentionally opinionated feature, intended to avoid over-namespacing and configuration anxiety. |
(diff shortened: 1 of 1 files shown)


10p-freddo:
Previous ideas and discussion:

- https://github.com/sveltejs/rfcs/pull/13
- https://github.com/sveltejs/rfcs/pull/22
- https://github.com/sveltejs/svelte/issues/901
- https://github.com/sveltejs/svelte/issues/2870
- https://github.com/sveltejs/svelte/issues/6422
- https://github.com/sveltejs/svelte/issues/6972

lukaszpolowczyk:
@ghughes See about the broader proposal "Targeted slots" (oddly similar name :O):

She's about Declarative Actions and Forward Directive, but it would work for passing styles too.
https://github.com/sveltejs/rfcs/pull/41#issuecomment-1126434719
https://github.com/sveltejs/rfcs/pull/60#issuecomment-1204651747

I recommend reading both threads, as I don't have one separate rfcs for this, and I discuss various details in both.


Targeted Styles and Targeted Slots are quite similar, but they do get rid of some of the problems you write about in your rfcs.
**Perhaps** it is possible to combine the two proposals. Still need to think about it, because I'm not sure.

Two people liked it. We'll see what happens next.

lukaszpolowczyk:
@ghughes Your example but using Targeted Slots:

```svelte
<!-- Parent.svelte -->

<div class="blue">
    <Child><svelte:element slot="button" class="button"/></Child>
</div>

<style>
/* Effective: */
.button {
    color: red;
}
.button span {
    font-family: "Comic Sans MS";
}

/* Ineffective, because `this="button"` is declared in Child and is not part of parent: */
.blue button {
    color: blue;
}
</style>
```

```svelte
<!-- Child.svelte -->

<!-- Will be styled: -->
<target targeted:button this="button" class="button">
    <span>Foo</span>
</target>

<!-- Will remain unstyled: -->
<button>Bar</button>

<!-- Allow the parent to style eligible elements in MenuItem: -->
<MenuItem inherit />
```

Note that I used `svelte:element` immediately closed, without the possibility of inserting content in parent.
And at the same time, instead of the `<slot>` tag, I gave the `<target></target>` tag, so that the content is not the replacement content, but the main content. This is the evolution of my idea about Targeted Slots.

It's nice that the class name can be the same in parent and child, because SvelteJS adds its hash (one in parent, one in child), so it won't interfere.

Unfortunately, I don't know what to do with `<MenuItem inherit />`. I guess that's what Targeted Slots can't do.
Maybe  `<MenuItem><svelte:element targeted:button/><MenuItem>`?

lukaszpolowczyk:
@ghughes Check out my RFCS:

Among other things, I write there about something that will solve almost all the problems from your RFCS.
There are examples.

lukaszpolowczyk:
@ghughes I just added changes that solve all the problems from your RFC.
Please check it out, and rate it! - https://github.com/sveltejs/rfcs/pull/68

AdaptingAFM:
have you thought about components having a default slot which is consumed when using a <style> block as their children? 

```svelte
/**
/ Parent Component
*/

<script lang='ts'>
  import Child from './Child.svelte';
</script>

<Child>
  <style lang='scss'>
    /* styles that would be added to the hoisted components' stylesheet — elements would be targeted & scoped directly by either using the components generated hash class or by adding a local style tag to the dom */

    div { background-color: green; }
  </style>
</Child>

<style lang='scss'>
  /* ...styles of the parent itself */
</style>

/**
/ Child Component
*/

<div>
  Hello!
</div>

<style lang='scss'>
  :where(div) { background-color: red; }
</style>
```

DylanYoung:
Maybe I'm missing something,  but isn't the solution to this and related issues already speced?

See https://developer.mozilla.org/en-US/docs/Web/CSS/::part

PR #69, opened 2022-11-13
Standardized Render Tag Syntax

ghost:
[Rendered](https://github.com/IRod22/svelte-rfcs/blob/master/text/0000-standardized-render-tag-syntax.md).

Related issues:
- sveltejs/svelte#3733

**Update:** I would prefer if participants commented instead of using :-1: reactions, so this RFC can get better with each comment

### DIFF
--- a/text/0000-standardized-render-tag-syntax.md
+++ b/text/0000-standardized-render-tag-syntax.md
@@ -1,4 +1,4 @@
-- Start Date: 2022-11-12
+- Start Date: 2022-11-13
 - RFC PR: (leave this empty)
 - Svelte Issue: (leave this empty)
 
@@ -9,8 +9,7 @@
 This RFC is a proposal for enhancing the existing control
 flow blocks (`{#if}`, `{#each}`, and `{#await}`) and tags
 (`{@html}` and `{@const}`), and add support for custom
-blocks and tags. The potential solution will use a developer
-friendly API, and the solution will use a standardized
+blocks and tags. The potential solution will use a standardized
 block and tag syntax.
 
 ## Motivation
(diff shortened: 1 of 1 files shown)


Machy8:
Hi + 1 from me, and I am adding inspiration from PHP World (please don't kill me).

Latte https://latte.nette.org/en/tags is using AST tree in his latest version. It provides features mentioned in this rfc `n:macro` and `{macro}`. Latte repo https://github.com/nette/latte.

It also has a guide for adding custom extensions https://latte.nette.org/en/creating-extension. 

This allows developers to create custom macros, filters, and functions. `N:macros` are used directly on elements and it can decrease the amount of code.

Extending macros allows developers to use macros like `sep`, `continueIf`, `breakIf`, `switch`, `capture`, `define`...

The combination of macros used directly on the element `s:macro` (for example) and classic macros, could lead to shorter, more readable and flexible templates. I am comparing it to Vue (only v:something) and Twig (which is also a PHP template engine but has fewer features than Latte and only provides {macros}, not attribute based macros).

Example with `s:macro` (svelte macro on the element).
```
<div s:inner-each="links as link" s:ifcontent>
  {#continueIf link.isExternal}
  <a href="{link.href}">{link.text}</a>
</div>
```
Normal macros (tags):
```
{#if links.length > 0}
  <div>
    {#each links as link}
        {#if !link.isExternal}
            <a href="{link.href}">{link.text}</a>
         {/if}
    {/each}
  </div>
{/if}

ghost:
> Hi + 1 from me, and I am adding inspiration from PHP World (please don't kill me).

Hello, @Machy8; thank you for your insights! I'll look into this this weekend!

ghost:
@Machy8 I just read the Latte Docs. I like the idea of extensions, but I'm wondering how they would be implement in both front-end and back-end environments. I have one idea for implementing the rendering logic: as seen in [this block of code](https://github.com/nimble-ui/core/blob/567b1c87395805c0186804956f7b8fdad72aadeb/index.ts#L89-L126) of a library I have been working on, a declarative approach could be implemented, but I think that could lower Svelte's efficiency.

ghost:
Latte-style extensions are a no-go... I just remembered that Svelte has to support Typescript. A different alternative I am thinking of would be like Vue's [custom directive](https://vuejs.org/guide/reusability/custom-directives.html#introduction) declarations. In other words, `{#if}` would correspond to a `svelteIf` function, `{@html}` with `svelteHtml`, and so-on.

